Issue #52, opened 2018-01-15
Invalidate cache when theme changed


asvishnyakov:
Version info: 
- Browser version: Chrome 63
- Platform version: 2.13.19
- Storefront version: 3.0.5

### Expected behavior
Storefront cache should be invalidated when theme files changed.

### Actual behavior
Nothing happens when Azure blob provider used. Locally working normal.

asvishnyakov:
@tatarincev @OlegoO We strongly need this feature, please try to make it within Operations as soon as possible

tatarincev:
released https://github.com/VirtoCommerce/vc-storefront-core/releases/edit/v3.0.7

PR #59, opened 2018-01-25
#52 Azure cache invalidate via pooling

ksavosteev:
Blob monitoring mostly taken from the old storefront. 
Added a custom implementation of IChangeToken - BlobChangeToken.
For each blob file BlobChangeToken is created and placed in ConcurrentDictionary (MonitorFileSystemChanges method).
Each iteration of monitor updates (if is already created) blob token's LastModified date so each token can have its own changed state.
After token obtains 'changed' state it's discarded and a new one is created (Watch method).


asvishnyakov:
@tatarincev Make review, please

tatarincev:
We have discussed and decided to make another implementation, but @ksavosteev is busy at the moment, and there is no one who could do it

VirtoCommerce.Storefront/appsettings.json
```diff
@@ -1,23 +1,27 @@
...
+            "RethrowLiquidRenderErrors": false
+        },
+        "RequireHttps": {
+            "Enabled": false,
+            "StatusCode": "308",
+            "Port": "443"
+        },
+        "AzureBlobStorage": {
+            "PollForChanges": true,
+            "PollingChangesInterval": "0:0:05"
```
tatarincev: Сould you rename this setting to `ChangesPoolingInterval` to be similar in the  `Virtocommerce:ChangesPoolingInterval` setting?

tatarincev:
Released https://github.com/VirtoCommerce/vc-storefront-core/releases/edit/v3.0.7

PR #66, opened 2018-02-07
Update FileSystemCountriesService.cs

HtrBr:
Use ISO2 as key for countries.json and add displayname


HtrBr:
ISO2 as key for countries.json

PR #72, opened 2018-02-13
Dev authentication for allowed stores

Andrew-Orlov:


VirtoCommerce.Storefront/Domain/Security/AuthWorkContextBuilderExtensions.cs
```diff
@@ -40,6 +40,12 @@ public static class AuthWorkContextBuilderExtensions
                 {
                     await signInManager.SignOutAsync();
                 }
+
+                //Current store is not allowed for signed in user - do sign out
+                if (!user.AllowedStores.Any(s => s.Equals(builder.WorkContext.CurrentStore.Id, StringComparison.OrdinalIgnoreCase)))
+                {
+                    await signInManager.SignOutAsync();
```
tatarincev: Missed  `user = null` assignment

VirtoCommerce.Storefront/Domain/Security/AuthWorkContextBuilderExtensions.cs
```diff
@@ -40,6 +40,12 @@ public static class AuthWorkContextBuilderExtensions
                 {
                     await signInManager.SignOutAsync();
                 }
+
+                //Current store is not allowed for signed in user - do sign out
+                if (!user.AllowedStores.Any(s => s.Equals(builder.WorkContext.CurrentStore.Id, StringComparison.OrdinalIgnoreCase)))
```
tatarincev: Need to replace  condition to this line user.AllowedStores.Contains(builder.WorkContext.CurrentStore)

PR #74, opened 2018-02-14
Update create user

n2pro:
as github writed user name in  custom claims

### DIFF
--- a/VirtoCommerce.Storefront/Controllers/AccountController.cs
+++ b/VirtoCommerce.Storefront/Controllers/AccountController.cs
@@ -243,8 +243,8 @@ namespace VirtoCommerce.Storefront.Controllers
                 await _signInManager.SignInAsync(user, isPersistent: false);
                 var registrationInfo = new UserRegistrationInfo
                 {
-                    FirstName = loginInfo.Principal.FindFirstValue(ClaimTypes.GivenName) ?? loginInfo.Principal.FindFirstValue(ClaimTypes.Name),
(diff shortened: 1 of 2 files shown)


VirtoCommerce.Storefront/Controllers/AccountController.cs
```diff
@@ -360,5 +359,17 @@ public ActionResult ForgotPassword()
                 return View("customers/account", WorkContext);
             }
         }
+
```
tatarincev: Better write the extension method for array calims names  `ClaimsPrincipal.FindFirstValue(string[] claimTypes, string defaultValue)` istead of made separate methods.

VirtoCommerce.Storefront/Controllers/AccountController.cs
```diff
@@ -360,5 +359,17 @@ public ActionResult ForgotPassword()
...
+        #region helpers methods for create user
+        private static string GetFirstName(ExternalLoginInfo loginInfo)
+        {
+            return loginInfo.Principal.FindFirstValue(ClaimTypes.GivenName) ?? loginInfo.Principal.FindFirstValue("urn:github:name") ?? loginInfo.Principal.FindFirstValue(ClaimTypes.Name) ?? "unknown";
+        }
+
+        private static string GetLastName(ExternalLoginInfo loginInfo)
```
tatarincev: Absolutely redundant method because there aren't repeatedly uses

PR #84, opened 2018-03-13
Dev wishlists search and create

ksavosteev:


VirtoCommerce.Storefront/Controllers/Api/ApiListsController.cs
```diff
@@ -107,15 +95,97 @@ public class ApiListsController : StorefrontControllerBase
...
+        [HttpPost]
+        public async Task<ActionResult> SearchLists([FromBody] ShoppingCartSearchCriteria searchCriteria)
+        {
+            if (searchCriteria == null)
+            {
+                searchCriteria = new ShoppingCartSearchCriteria();
+            }
+
+            //restricting query to lists search, only paging and sorting are available
+            searchCriteria.Type = "wishlist";
```
asvishnyakov: @tatarincev Would it be better to use "wishlist" as default parameter of search criteria, instead of hard coding? Possible will allow us to use multiple types of lists (i.e. wishlist, recipe, something else...) without changing code.

VirtoCommerce.Storefront.Model/Lists/Services/IWishlistService.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Threading.Tasks;
+using PagedList.Core;
+using VirtoCommerce.Storefront.Model.Lists;
+
+namespace VirtoCommerce.Storefront.Model.Lists.Services
+{
+    public interface IWishlistService
+    {
+        Task<IPagedList<Wishlist>> SearchShoppingCartsAsync(WishlistSearchCriteria criteria);
```
tatarincev: Need  rename to SearchWishListsAsync

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -24,11 +26,10 @@
 using VirtoCommerce.Storefront.Model.Stores;
 using VirtoCommerce.Storefront.Model.Subscriptions.Services;
 using VirtoCommerce.Storefront.Model.Tax.Services;
-using cartModel = VirtoCommerce.Storefront.AutoRestClients.CartModuleApi.Models;
 
 namespace VirtoCommerce.Storefront.Domain
 {
-    public class CartBuilder : ICartBuilder
+    public class CartBuilder : IWishlistBuilder
```
tatarincev: Need to inherit from ICartBuilder

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -37,9 +38,10 @@ public class CartBuilder : ICartBuilder
         private readonly IPromotionEvaluator _promotionEvaluator;
         private readonly ITaxEvaluator _taxEvaluator;
         private readonly ISubscriptionService _subscriptionService;
+        public readonly IWishlistService _listService;
 
         public CartBuilder(IWorkContextAccessor workContextAccessor, ICartModule cartApi, ICatalogService catalogSearchService,
-            IMemoryCache memoryCache, IPromotionEvaluator promotionEvaluator, ITaxEvaluator taxEvaluator, ISubscriptionService subscriptionService)
+            IMemoryCache memoryCache, IPromotionEvaluator promotionEvaluator, ITaxEvaluator taxEvaluator, ISubscriptionService subscriptionService, IWishlistService listService)
```
tatarincev: Why this there?

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -449,18 +450,20 @@ public virtual Task RemoveShipmentAsync(string shipmentId)
 
         #endregion      
 
-        protected virtual cartModel.ShoppingCartSearchCriteria CreateCartSearchCriteria(string cartName, Store store, User user, Language language, Currency currency)
+        protected virtual WishlistSearchCriteria CreateCartSearchCriteria(string cartName, Store store, User user, Language language, Currency currency, string type)
```
tatarincev: CardBuilder hasn't to know anything about Wishlists, need to change back to ShoppingCartSearchCriteria

VirtoCommerce.Storefront/Domain/Lists/WishlistConverter.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        public static cartDto.ShoppingCartSearchCriteria ToSearchCriteriaDto(this WishlistSearchCriteria criteria)
+        {
+            return WishlistConverterInstance.ToSearchCriteriaDto(criteria);
+        }
+    }
+
+    public partial class WishlistConverter : CartConverter
+    {
```
tatarincev: I would  suggest use there a hack with Json.Convert to convert base class (Cart) to derived class (Wishlist) instead of this assignments

VirtoCommerce.Storefront.Model/Lists/Services/IWishlistService.cs
asvishnyakov: @tatarincev Don't use wish- prefix please, it's lists, not wishlists only

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
ksavosteev: fixed

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
ksavosteev: removed

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
ksavosteev: fixed

VirtoCommerce.Storefront/Domain/Lists/WishlistConverter.cs
ksavosteev: I'd rather not - it's a bit slow, need to add many constructors to cart, cart line item, etc

VirtoCommerce.Storefront.Model/Lists/Services/IWishlistService.cs
ksavosteev: since the main object is called "Wishlist" already shouldn't all utility services have this prefix?

asvishnyakov: @ksavosteev it's not about specific place. yes, if we call it wishlists, then we should use it everywhere and you may ignore my comment. but I think it's not about wishlists only

VirtoCommerce.Storefront/Controllers/Api/ApiListsController.cs
```diff
@@ -93,29 +82,110 @@ public class ApiListsController : StorefrontControllerBase
...
+                searchCriteria = new WishlistSearchCriteria();
+            }
+
+            //restricting query to lists search, only paging and sorting are available
+            searchCriteria.StoreId = WorkContext.CurrentStore.Id;
+            searchCriteria.Customer = WorkContext.CurrentUser;
+            searchCriteria.Currency = WorkContext.CurrentCurrency;
+            searchCriteria.Language = WorkContext.CurrentLanguage;
+
+            var cartPagedList = await _listService.SearchWishlistsAsync(searchCriteria);
```
tatarincev: need to remove comment

VirtoCommerce.Storefront.Model/Lists/Services/IWishlistService.cs
Woland2k: @ksavosteev @asvishnyakov I agree, it should be just lists, not specific wish lists

PR #90, opened 2018-03-21
Added support user lockout

n2pro:
Since we do not have a UserManager implementation on storefront , but when _signInManager.PasswordSignInAsync is called [this](https://github.com/aspnet/Identity/blob/c6a82ad19a0bdb80b7358dc843b66664bbf33f96/src/Identity/SignInManager.cs#L273),  then [this](https://github.com/aspnet/Identity/blob/c6a82ad19a0bdb80b7358dc843b66664bbf33f96/src/Identity/SignInManager.cs#L319), then [this](https://github.com/aspnet/Identity/blob/c6a82ad19a0bdb80b7358dc843b66664bbf33f96/src/Identity/SignInManager.cs#L786), then [this](https://github.com/aspnet/Identity/blob/c6a82ad19a0bdb80b7358dc843b66664bbf33f96/src/Identity/SignInManager.cs#L765), then [this](https://github.com/aspnet/Identity/blob/9ecbefcf21d12d90ac187e1eefe02ed0219edc23/src/Core/UserManager.cs#L1898), IUserLockoutStore<User> must be realized with the implementation of these methods


tatarincev:
I've already implemented this feature in another way.
see details in this commit https://github.com/VirtoCommerce/vc-storefront-core/commit/7f56992fa8287cb68c6f843319a9e783a7cf7628

PR #103, opened 2018-03-29
Added derivative contracts support

eMazeika:


asvishnyakov:
@tatarincev Please, review. Changes was made in view of the fact that storefront may be used without DerivativeContracts module installed

PR #114, opened 2018-05-31
Fixed mapping BlogSearch to liquid

n2pro:


n2pro:
image
Lost mapping members after migrating to storefront core.

PR #121, opened 2018-06-14
impersonating: redirect Anonymous to login page

eMazeika:
impersonating: redirect Anonymous to login page

### DIFF
--- a/VirtoCommerce.Storefront/Controllers/AccountController.cs
+++ b/VirtoCommerce.Storefront/Controllers/AccountController.cs
@@ -196,6 +196,11 @@ namespace VirtoCommerce.Storefront.Controllers
 
         public async Task<IActionResult> ImpersonateUser(string userId)
         {
+            if (User.Identity.Name == SecurityConstants.AnonymousUsername)
+            {
+                return StoreFrontRedirect($"~/account/login?ReturnUrl={Request.Path}");
+            }
+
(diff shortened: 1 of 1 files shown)


VirtoCommerce.Storefront/Controllers/AccountController.cs
```diff
@@ -196,6 +196,11 @@ public ActionResult ConfirmInvitation(string organizationId, string email, strin
 
         public async Task<IActionResult> ImpersonateUser(string userId)
         {
+            if (User.Identity.Name == SecurityConstants.AnonymousUsername)
+            {
+                return StoreFrontRedirect($"~/account/login?ReturnUrl={Request.Path}");
```
artem-dudarev: Request.Path should be URL-encoded

VirtoCommerce.Storefront/Controllers/AccountController.cs
```diff
@@ -196,6 +197,11 @@ public ActionResult ConfirmInvitation(string organizationId, string email, strin
 
         public async Task<IActionResult> ImpersonateUser(string userId)
         {
+            if (User.Identity.Name == SecurityConstants.AnonymousUsername)
+            {
+                return StoreFrontRedirect($"~/account/login?ReturnUrl={HttpUtility.UrlEncode(Request.Path)}");
```
tatarincev: You should avoid of use HttpUtility in asp.net core projects use `IUrlEncoder`  dependency instead. And in this case, encoding absolutely redundant

PR #128, opened 2018-07-04
Redirect to login when cookie expired

asvishnyakov:
By default, ASP.NET Core do nothing after user cookie expiration. So:
* After cookie expiration customer may have bad UX: for example, they loose their cart
* If user was on page which require permissions, when they will be redirected to "Access denied" page

To prevent this, I want to redirect customer to login page when cookie expires.


artem-dudarev:
Also, when building the ligin URL, we should take into account the store name.
For example, currently the login URL is always `/account/login`, but in some cases it should be `/{store}/{language}/account/login`

PR #132, opened 2018-07-05
Update Money.cs

HtrBr:
Quick fix , no decimals for higher prices.


tatarincev:
Could you repeat this PR to  the dev branch?

asvishnyakov:
@HtrBr I believe @tatarincev mean "please, change target branch to dev"

artem-dudarev:
@HtrBr The commit says `no decimals for higher prices`, but the code removes group separator, and there is no check for `higher prices` - the separator is always removed.
What this PR is about?

PR #139, opened 2018-07-31
Descriptions without language used as fallback https://github.com/Vir…

eMazeika:
https://github.com/VirtoCommerce/vc-module-catalog/issues/107


VirtoCommerce.Storefront/Domain/Catalog/CatalogConverter.cs
```diff
@@ -439,12 +439,22 @@ public static Product ToProduct(this catalogDto.Product productDto, Language cur
 
             if (productDto.Reviews != null)
             {
-                result.Descriptions = productDto.Reviews.Where(r => !string.IsNullOrEmpty(r.Content)).Select(r => new EditorialReview
+                var productReviews = productDto.Reviews
```
tatarincev: There should be a grouping for `ReviewType`  and logic for the language

Issue #138, opened 2018-07-31
Eliminate manual route registration


tatarincev:
Currently, all storefront routes are registering through manually adding lines 
 in the extension https://github.com/VirtoCommerce/vc-storefront-core/blob/master/VirtoCommerce.Storefront/Routing/MapStorefrontRouteBuilderExtension.cs#L18
That's actions are necessary because of storefront routing has the next format /{storeId}/{locale}/ and we must to normalized each request before the routing process.
https://virtocommerce.com/docs/vc2devguide/working-with-storefront/storefront-seo-routing

Need to find out a solution to eliminate these manual route registration and use only  ASP.NET Core standard routes registration way  (Attributes or Controller and methods names)

avas:
Vitaly added some routing tests while working on this task. I modified these tests to test actual routing and added test case for all manual routing entries. After that, I just used `StorefrontRouteAttribute` and `StorefrontApiRouteAttribute` added by Vitaly to create same routes and eliminate manual routing setup.

avas:
After transition to attribute-based routing dynamic slug routes stopped working. This problem is caused by the fact that ASP.NET Core MVC uses 2 different routers:
* `AttributeRoute` that handles attribute-based routes (executes before `SlugRoute`);
* `MvcRouteHandler` that handles manually registered routes (acts as handler for `Route` and executes right after `SlugRoute.OnRouteMatched`).

Usually MVC route contains some specific implementation of `IRouter` that attempts to find matching action and, if succeeded, sets request handler by assigning value to `RouteContext.Handler`.  `SlugRoute` is derived from `Route` and uses `MvcRouteHandler` for that purpose, but it doesn't set `Handler` explicitly - instead, it just gives some hints to `MvcRouteHandler` by filling controller name, action name and request parameters to `RouteContext.RouteData.Values`. Previously, when routes were registered manually, `MvcRouteHandler` handled all controller actions, so it was correctly getting controller action by those hints. But now, after transition to attribute-based routing, all actions are handled by `AttributeRoute`, and `MvcRouteHandler` handles nothing. That's why it couldn't resolve controller action, and for dynamic routes storefront was returning 404.

So, to solve this problem, I modified `SlugRoute`. Now it attempts to find matching controller action itself and, if succeeded, assigns `RouteContext.Handler` that invokes that action. This way dynamic slug routes appear to work correctly.

tatarincev:
[...]

PR #153, opened 2018-08-31
138: eliminating manual route registration

avas:
Changes made for #138:
* Added tests for all manual routes;
* Replaced manual routing with attribute-based routing and verified (by running tests) that it still works the same way.

Routing tests (`RoutingTests.TestRouting`) do the following:
* Start an in-memory web server that handles request;
* Override authorization by adding `AllowAnonymousFilter`;
* Disable antiforgery by adding `DummyAntiforgery` instead of default implementation;
* Most crucial part - add the `RoutingTestingActionFilter`. It prevents running the actual target action - instead, it returns `RoutingDataResult` with some information about routing (i.e. what was requested, which controller method would process that request and what arguments will be passed to it);
* Configure rest of required parameters using `Startup` class from `VirtoCommerce.Storefront`;
* Send requests to target URL and verify routing data from received `RoutingDataResult`. Each request gets sent to 4 addresses:
    * raw target URL;
    * `$"en-US/{targetURL}"`;
    * `$"Electronics/{targetURL}"`;
    * `$"Electronics/en-US/{targetURL}"`.


avas:
Sorry, I thought it is just another manual route and didn't know it was so important. Restored it back.

PR #155, opened 2018-09-05
Many coupons implementation

andrey-ciclicci:
Extended [ShoppingCart](https://github.com/VirtoCommerce/vc-storefront-core/blob/master/VirtoCommerce.Storefront.Model/Cart/ShoppingCart.cs), [CartConverter](https://github.com/VirtoCommerce/vc-storefront-core/blob/master/VirtoCommerce.Storefront/Domain/Cart/CartConverter.cs), [MarketingConverter ](https://github.com/VirtoCommerce/vc-storefront-core/blob/master/VirtoCommerce.Storefront/Domain/Marketing/MarketingConverter.cs#L108)and [CartBuilder](https://github.com/VirtoCommerce/vc-storefront-core/blob/master/VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs#L28) to support multiple coupons.


VirtoCommerce.Storefront.Model/Cart/ShoppingCart.cs
```diff
@@ -363,11 +372,11 @@ public void ApplyRewards(IEnumerable<PromotionReward> rewards)
                 payment.ApplyRewards(paymentRewards);
             }
 
-            if (Coupon != null && !string.IsNullOrEmpty(Coupon.Code))
+            foreach (var coupon in Coupons)
             {
-                Coupon.AppliedSuccessfully = rewards.Any(x => x.IsValid && x.Coupon != null);
+                if (!string.IsNullOrEmpty(coupon.Code))
```
tatarincev: Code can be simplified
```
coupon.AppliedSuccessfully = !string.IsNullOrEmpty(coupon.Code) && rewards.Any(x => x.IsValid && x.Coupon.EqualsInvariant(coupon.Code));
```

VirtoCommerce.Storefront/Controllers/Api/ApiCartController.cs
```diff
@@ -224,25 +224,25 @@ public async Task<ActionResult> AddCartCoupon(string couponCode)
             {
                 var cartBuilder = await LoadOrCreateCartAsync();
 
-                await cartBuilder.AddCouponAsync(couponCode);
+                var isUnique = await cartBuilder.AddCouponAsync(couponCode);
```
tatarincev: Could you clarify why do we need this kind of check on this level?

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -168,17 +168,26 @@ public virtual Task RemoveItemAsync(string id)
             return Task.FromResult((object)null);
         }
 
-        public virtual Task AddCouponAsync(string couponCode)
+        public virtual Task<bool> AddCouponAsync(string couponCode)
```
tatarincev: We don't need return bool value from this method. The code duplication check can be performed on the UI layer or a distinct addition to the cart.coupons collection

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -168,17 +168,26 @@ public virtual Task RemoveItemAsync(string id)
...
+                return Task.FromResult(false);
+            }
         }
 
-        public virtual Task RemoveCouponAsync()
+        public virtual Task RemoveCouponAsync(string couponCode)
         {
             EnsureCartExists();
-            Cart.Coupon = null;
+            var coupon = Cart.Coupons.FirstOrDefault(c => c.Code == couponCode);
```
tatarincev: You can use the `Cart.Coupons.RemoveAll` instead of these lines

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -474,7 +483,8 @@ protected virtual ShoppingCart CreateCart(string cartName, Store store, User use
                 Customer = user,
                 Type = type,
                 IsAnonymous = !user.IsRegisteredUser,
-                CustomerName = user.IsRegisteredUser ? user.UserName : SecurityConstants.AnonymousUsername
+                CustomerName = user.IsRegisteredUser ? user.UserName : SecurityConstants.AnonymousUsername,
+                Coupons = new List<Coupon>(),
```
tatarincev: Move this initialization to the ShoppingCart constructor

VirtoCommerce.Storefront/Domain/Cart/CartConverter.cs
```diff
@@ -522,7 +522,7 @@ public static cartDto.ShoppingCart ToShoppingCartDto(this ShoppingCart cart)
                 result.LanguageCode = cart.Language.CultureName;
             }
             result.Addresses = cart.Addresses.Select(ToCartAddressDto).ToList();
-            result.Coupon = cart.Coupon != null ? cart.Coupon.Code : null;
+            result.Coupons = cart.Coupons != null ? cart.Coupons.Select(c => c.Code).ToList() : null;
```
tatarincev: This line won't be compiled isn't?

VirtoCommerce.Storefront/Controllers/Api/ApiCartController.cs
andrey-ciclicci: The idea was not to rely on front-end logic and to avoid returning wrong object when a duplicate coupon is filtered out by the "AddCouponAsync" check.
Currently the return value is not being used at all in default theme ("GetCart" is called), should we always return "Json(null)" instead?

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
andrey-ciclicci: Is it ok to convert it using .ToList() first?

VirtoCommerce.Storefront/Domain/Cart/CartConverter.cs
andrey-ciclicci: The "Coupons" field of "cartDto.ShoppingCart" object has a list of strings type:
   public IList<string> Coupons { get; set; }
Dto cart object is being sent to "api/carts" to save the cart.

artem-dudarev: Can be simplified:
`result.Coupons = cart.Coupons?.Select(c => c.Code).ToList();`

VirtoCommerce.Storefront/Domain/Cart/CartConverter.cs
```diff
@@ -429,13 +429,13 @@ public static ShoppingCart ToShoppingCart(this cartDto.ShoppingCart cartDto, Cur
                 Customer = user
             };
 
-            if (cartDto.Coupon != null)
+            if (cartDto.Coupons != null)
             {
-                result.Coupon = new Coupon
+                result.Coupons = new List<Coupon>(cartDto.Coupons.Select(coupon => new Coupon
```
artem-dudarev: Can be simplified:
`result.Coupons = cartDto.Coupons.Select(coupon => new Coupon { Code = coupon, AppliedSuccessfully = !string.IsNullOrEmpty(coupon) }).ToList();`

VirtoCommerce.Storefront/Domain/Marketing/MarketingConverter.cs
```diff
@@ -114,7 +114,7 @@ public static marketingDto.PromotionEvaluationContext ToPromotionEvaluationConte
                 result.CartPromoEntries = promoEvalContext.Cart.Items.Select(x => x.ToProductPromoEntryDto()).ToList();
 
                 result.CartTotal = (double)promoEvalContext.Cart.SubTotal.Amount;
-                result.Coupon = promoEvalContext.Cart.Coupon != null ? promoEvalContext.Cart.Coupon.Code : null;
+                result.Coupons = promoEvalContext.Cart.Coupons != null ? promoEvalContext.Cart.Coupons.Select(c => c.Code).ToList() : null;
```
artem-dudarev: result.Coupons = promoEvalContext.Cart.Coupons?.Select(c => c.Code).ToList();

VirtoCommerce.Storefront/Domain/Cart/CartConverter.cs
andrey-ciclicci: Followed the same code pattern as for the other 'cartDto' items below. Should we keep it or refactor the population of other lists too?

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -171,14 +171,18 @@ public virtual Task RemoveItemAsync(string id)
         public virtual Task AddCouponAsync(string couponCode)
         {
             EnsureCartExists();
-            Cart.Coupon = new Coupon { Code = couponCode };
+            if (!Cart.Coupons.Any(c => c.Code == couponCode))
```
artem-dudarev: Use case insensitive comparison: `c.Code.EqualsInvariant(couponCode)`

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -171,14 +171,18 @@ public virtual Task RemoveItemAsync(string id)
...
+
             return Task.FromResult((object)null);
         }
 
-        public virtual Task RemoveCouponAsync()
+        public virtual Task RemoveCouponAsync(string couponCode)
         {
             EnsureCartExists();
-            Cart.Coupon = null;
+            Cart.Coupons.ToList().RemoveAll(c => c.Code == couponCode);
```
artem-dudarev: Use case insensitive comparison

VirtoCommerce.Storefront/Domain/Cart/CartConverter.cs
```diff
@@ -429,13 +429,9 @@ public static ShoppingCart ToShoppingCart(this cartDto.ShoppingCart cartDto, Cur
...
 
-            if (cartDto.Coupon != null)
+            if (cartDto.Coupons != null)
             {
-                result.Coupon = new Coupon
-                {
-                    Code = cartDto.Coupon,
-                    AppliedSuccessfully = !string.IsNullOrEmpty(cartDto.Coupon)
-                };
+                result.Coupons = cartDto.Coupons?.Select(c => new Coupon { Code = c, AppliedSuccessfully = !string.IsNullOrEmpty(c) }).ToList();
```
artem-dudarev: You have checked `cartDto.Coupons` for null already, so you don't need to use `?.`

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -171,14 +171,18 @@ public virtual Task RemoveItemAsync(string id)
...
+
             return Task.FromResult((object)null);
         }
 
-        public virtual Task RemoveCouponAsync()
+        public virtual Task RemoveCouponAsync(string couponCode)
         {
             EnsureCartExists();
-            Cart.Coupon = null;
+            Cart.Coupons.Remove(Cart.Coupons.FirstOrDefault(c => c.Code == couponCode));
```
artem-dudarev: Use case insensitive comparison

PR #183, opened 2018-12-21
#49: "For N in every Y…" rewards storefront support

yecli:
- Added "For N in every Y…" for one product to LineItem and Product, so they are properly applied and calculated on storefront.
- Added tests for LineItem  reward calculation and Product reward appliance.


VirtoCommerce.Storefront.Model/Cart/LineItem.cs
```diff
@@ -312,12 +312,29 @@ public void ApplyRewards(IEnumerable<PromotionReward> rewards)
...
+                return;
+            }
+
             foreach (var reward in lineItemRewards)
             {
                 var discount = reward.ToDiscountModel(ListPrice - DiscountAmount);
                 if (reward.Quantity > 0)
                 {
-                    discount.Amount = discount.Amount * Math.Min(reward.Quantity, Quantity) / Quantity;
+                    //If we have condition like For "N in every Y …"
```
tatarincev: You must encapsulate the implementation of  this new logic in the PromotionReward.ToDiscountModel() method. 
And please take a already exists implementation from here https://github.com/VirtoCommerce/vc-module-core/blob/dev/VirtoCommerce.Domain/Marketing/Model/Promotions/Rewards/AmountBasedReward.cs#L48

yecli: Done. Please note that logic from vc-core-module AmountBasedReward does not contain storefront feature that MaxLimit works only on AmountType.Relative rewards (that's why i did not touch this method before - to not break existing storefront rewards logic for all items). Seems now it is not actual, and storefront calculation logic is the same as in vc-core-module AmountBasedReward. Also removed MarkertingTests that checked this AmountType.Relative feature.

VirtoCommerce.Storefront.Model/Catalog/Product.cs
```diff
@@ -389,7 +389,9 @@ public void ApplyTaxRates(IEnumerable<TaxRate> taxRates)
 
         public void ApplyRewards(IEnumerable<PromotionReward> rewards)
         {
-            var productRewards = rewards.Where(r => r.RewardType == PromotionRewardType.CatalogItemAmountReward && (r.ProductId.IsNullOrEmpty() || r.ProductId.EqualsInvariant(Id)));
+            var productRewards = rewards.Where(r => r.RewardType == PromotionRewardType.CatalogItemAmountReward && (r.ProductId.IsNullOrEmpty() || r.ProductId.EqualsInvariant(Id))
```
tatarincev: We do not need these rules for the product, because there is no quantity

yecli: If we have condition "Get 50% off for 1 product1 for every 1 of product1" or "Get 50% off for 1 product1 for every 1 of product2" - don't we need to show discounted product price (for 1 entry, on page on screenshot)?
image

yecli: Fixed.

tatarincev:
Release  https://github.com/VirtoCommerce/vc-storefront-core/releases/v3.2.7

Issue #186, opened 2018-12-27
Need to limit the PageSize taken from the query string 


tatarincev:
The problem description:
Can a user generate a URL with a follow query string `?page_size = biginteger`, that parameter can lead to a large amount of data loading and may negative affect to application performance.

Technical solution:
Need to add a special setting that will contain the maximum page size that can be taken from the query string and use this limitation in here code
https://github.com/VirtoCommerce/vc-storefront-core/blob/dev/VirtoCommerce.Storefront/Domain/WorkContextBuilder.cs#L22

tatarincev:
Released https://github.com/VirtoCommerce/vc-storefront-core/releases/edit/v3.2.7

PR #187, opened 2018-12-27
#186 Maximum value for page size taken from query string

Andrew-Orlov:


Woland2k:
There should be a limit to the maximum the page size can be, for instance 100, so malicious user can't request 10000 product at once and crash the system.

Issue #51, opened 2018-01-12
IIS Url Rewrite role doesn't work


OlegoO:
### Expected behavior
See staging content from robots.staging.tx

### Actual behavior
See main robots.txt

### Steps to reproduce
1. Add rule 
````````````
<rule name="Robots.txt for production domain only" stopProcessing="true">
            <match url="robots.txt" />
            <conditions>
                <add input="{HTTP_HOST}" pattern="^(virtocommerce\.com)$" negate="true"/>
            </conditions>
            <action type="Rewrite" url="/robots.staging.txt" />
        </rule>
````````````

2. Open https://vc-public-qc.azurewebsites.net/robots.txt
3. Should see content from robots.staging.txt

**Comment by tatarincev** [comment]
This rule doesn't work because ASP.NET Static files middleware handle exists files first rather than UrlRewrite due to the registration order in Startup.cs 
```
...
  app.UseStaticFiles();
...
 app.UseRewriter(rewriteOptions);
```

So my suggestion to define two `production-robots.txt` and `staging-robots.txt` files and add to IISUrlRewrite.xml next rules
```
 <rule name="Robots.txt prod" stopProcessing="true">
            <match url="robots.txt" />
            <conditions>
                <add input="{HTTP_HOST}" pattern="^(virtocommerce\.com)$" negate="true"/>
            </conditions>
            <action type="Rewrite" url="/staging-robots.txt" />
        </rule>
        <rule name="Robots.txt staging" stopProcessing="true">
            <match url="robots.txt" />
            <conditions>
                <add input="{HTTP_HOST}" pattern="^(virtocommerce\.com)$"/>
            </conditions>
            <action type="Rewrite" url="/production-robots.txt" />
        </rule>
```

PR #189, opened 2019-01-11
#51 Fixed shopper gender issue

Andrew-Orlov:


tatarincev:
I've made another solution based on UserGroups

PR #192, opened 2019-01-23
Fix for Money.TruncatedAmount behavior

avas:
I recently noticed a strange behavior of the `Money.TruncatedAmount` property. According to the description, it should return the value of `InternalAmount` truncated by `DecimalDigits`, but it sometimes doesn't do that. For example, if the `InternalAmount` is equal to `0.999` and `DecimalDigits` is equal to 2, I expect it to return `0.99`, but instead it returns `0.5`.
So, I've attempted to fix the `TruncatedAmount` property and add some tests for it. What do you think? Have I got the purpose of this property right?

### DIFF
--- a/VirtoCommerce.Storefront.Tests/Model/MoneyOperationTests.cs
+++ b/VirtoCommerce.Storefront.Tests/Model/MoneyOperationTests.cs
@@ -4,6 +4,7 @@ using StorefrontLanguage = VirtoCommerce.Storefront.Model.Language;
 
 namespace VirtoCommerce.Storefront.Tests.Model
 {
+    [Trait("Category", "CI")]
     public class MoneyOperationTests
     {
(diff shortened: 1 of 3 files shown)


VirtoCommerce.Storefront.Tests/Model/MoneyOperationTests.cs
```diff
@@ -0,0 +1,40 @@
+using VirtoCommerce.Storefront.Model.Common;
+using Xunit;
+using StorefrontLanguage = VirtoCommerce.Storefront.Model.Language;
+
+namespace VirtoCommerce.Storefront.Tests.Model
+{
+    public class MoneyOperationTests
```
tatarincev: Could you add Traits("CI")  attribute for test class?

avas: Sorry, I forgot about that attribute. Added it.

PR #200, opened 2019-01-31
Add budget and it validation

AkulovOleg:
Add budget field to Contact, so if we decide to add new column in database we need to change only Budget property in Contact model

### DIFF
--- a/VirtoCommerce.Storefront.Model/Security/OrganizationUserRegistration.cs
+++ b/VirtoCommerce.Storefront.Model/Security/OrganizationUserRegistration.cs
@@ -9,9 +9,8 @@ namespace VirtoCommerce.Storefront.Model.Security
         public string Role { get; set; }
 
         [FromForm(Name = "customer[budget]")]
-        [Required]
         [Range(typeof(decimal), "0", "79228162514264337593543950335")]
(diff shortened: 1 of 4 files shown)


VirtoCommerce.Storefront.Model/Security/OrganizationUserRegistration.cs
```diff
@@ -1,11 +1,18 @@
...
 
 namespace VirtoCommerce.Storefront.Model.Security
 {
     public partial class OrganizationUserRegistration : UserRegistration
     {
         [Required]
         public string Role { get; set; }
+
+        [FromForm(Name = "customer[budget]")]
+        [Required]
```
vashaleev: Remove attribute Required
Set decimal is nullable

VirtoCommerce.Storefront.Model/Security/OrganizationUserRegistration.cs
```diff
@@ -1,11 +1,18 @@
...
+        [Range(typeof(decimal), "0", "79228162514264337593543950335")]
+        public decimal Budget { get; set; }
```
vashaleev: Set decimal is nullable

PR #204, opened 2019-02-12
Feature 197 2fa login workflow

yecli:
Added two factor authentication into login workflow.


VirtoCommerce.Storefront/Controllers/AccountController.cs
```diff
@@ -566,35 +567,76 @@ public async Task<ActionResult> ForgotLogin(ForgotPassword formModel)
...
+            var user = await _signInManager.UserManager.FindByEmailAsync(formModel.Email);
             if (user == null)
             {
                 WorkContext.ErrorMessage = "User was not found.";
                 return View("error", WorkContext);
             }
 
-            var isValidToken = false;
+            var phoneNumber = await _signInManager.UserManager.GetPhoneNumberAsync(user);
+            var isValidToken = await _signInManager.UserManager.VerifyChangePhoneNumberTokenAsync(user, formModel.Code, phoneNumber);
```
artem-dudarev: It's not a _change phone number_ token, it's a _reset password token_

yecli: We are using ChangePhoneNumberToken intentionally (6 digits code) as ResetPasswordToken generates token that is good to send by email as a part of link (like http://localhost:2082/Electronics/en-US/account/resetpassword?UserId=x&Token=**CfDJ8H9Z13d0HyxJviWis5gewrrlLU5bCgtuZo8gmfCWpQXflXw2RLFxV4crrbaB05Z3UdSNVnXQpVULvpcEtHmj6ayHfN%2FQaizGMKBVj0Qu054oQfd2Q7EvSnJDHfqKVQ%2BsnhaGl8c9qEOZ0MqpiixDNHPQ2QcLxbxt%2BaLickWNAqx5avPvPFOuuwphTmYpCYJ2AHPEZLda9iM8bJpblDOEDDZMNzJ1UWn%2BgSVhXG9uPBzyfT9fL0DEpmon3qCmPqTxbQ%3D%3D**).
There is no ResetPasswordToken suitable for sms sending in ASP NET Core Indentity.

yecli: Also please note this is not latest commit for review, as SMS sending part will be rework using Sms notifications implemented on platform level.

artem-dudarev: The problem is that the purpose of this token is to change the phone number:
```csharp
    public virtual Task<string> GenerateChangePhoneNumberTokenAsync(TUser user, string phoneNumber)
    {
      ThrowIfDisposed();
      return GenerateUserTokenAsync(user, Options.Tokens.ChangePhoneNumberTokenProvider, "ChangePhoneNumber:" + phoneNumber);
    }
```

Here is more correct way:
```csharp
var token = await _signInManager.UserManager.GenerateUserTokenAsync(user, TokenOptions.DefaultPhoneProvider, "ResetPassword");
var isValidToken = await _signInManager.UserManager.VerifyUserTokenAsync(user, TokenOptions.DefaultPhoneProvider, "ResetPassword", token);
```

yecli: Changed, thanks for the help!

VirtoCommerce.Storefront/Controllers/AccountController.cs
```diff
@@ -285,14 +289,79 @@ public async Task<ActionResult> Login([FromForm] Login login, string returnUrl)
...
+                    WorkContext.ErrorMessage = $"Operation failed";
+                    return View("error", WorkContext);
+                }
+
+                // Generate the token and send it
+
+                // TODO: Support configurable providers ("Phone/Email")
+                // https://github.com/aspnet/Docs/blob/aac9bf5e4aaa4f593648ddb7b92ff0e52f2f6a47/aspnetcore/security/authentication/2fa/sample/Web2FA/Controllers/AccountController.cs#L380
+                // Now we are using only Phone provider
+                var selectedProvider = "Phone";
```
tatarincev: Why is here "phone" provider hard-coded? Shouldn't it been taken from settings?

VirtoCommerce.Storefront/Controllers/Api/ApiAccountController.cs
```diff
@@ -424,5 +424,27 @@ public async Task<ActionResult> UpdateAddresses([FromBody] IList<Address> addres
 
             return Ok();
         }
+
+        // DELETE: storefrontapi/account/phonenumber
+        [HttpDelete("phonenumber")]
```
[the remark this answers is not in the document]
yecli: Done

VirtoCommerce.Storefront/VirtoCommerce.Storefront.csproj
```diff
@@ -36,6 +36,7 @@
   <ItemGroup>
     <PackageReference Include="AspNet.Security.OAuth.GitHub" Version="2.0.0" />
     <PackageReference Include="AspNet.Security.OAuth.StackExchange" Version="2.0.0" />
+    <PackageReference Include="ASPSMS" Version="0.0.2" />
```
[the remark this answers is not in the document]
yecli: No changes in csproj anymore

VirtoCommerce.Storefront/Controllers/AccountController.cs
yecli: Fixed

VirtoCommerce.Storefront/Controllers/Api/ApiAccountController.cs
```diff
@@ -430,5 +431,59 @@ public async Task<ActionResult> UpdateAddresses([FromBody] IList<Address> addres
...
+        {        
+            var phoneConfirmed = await _signInManager.UserManager.IsPhoneNumberConfirmedAsync(WorkContext.CurrentUser);
+            if (!phoneConfirmed)
+            {
+                var url = "/account/phonenumber";
+
+                return new ChangeTwoFactorAuthenticationResult { Succeeded = false, VerificationUrl = url };
+            }
+                             
+            var result = await _signInManager.UserManager.SetTwoFactorEnabledAsync(WorkContext.CurrentUser, !model.Enabled);
```
yecli: Why **!** model.Enabled (NOT is the problem) - could we use value we get from model? Probably we could change it when passing from JS.

ilyawzrd:
@yecli 
**TwoFactorAuth workflow**
TwoFactorAuth

**Update/Delete workflow**
updateDelete

VirtoCommerce.Storefront/appsettings.Development.json
```diff
@@ -8,7 +8,11 @@
     }
   },
   "VirtoCommerce": {
-    "Endpoint": {    
+    "CacheEnabled": "false",
```
yecli: Probably we do not need it in Dev config, as such settings are specific to the user (e.g. Url and SecretKey). User Secrets is the place when we need to specify local settings.
image
So I think it's better to revert this file.

trueboroda: ok

PR #207, opened 2019-02-12
Replace pooling with polling

artem-dudarev:


yecli:
Released in [v3.2.9](https://github.com/VirtoCommerce/vc-storefront-core/releases/tag/v3.2.9)

PR #228, opened 2019-03-25
fix generic type for swagger 

AliveMen:
get rid of generic result type


VirtoCommerce.Storefront.Model/Common/SearchResults/CustomerOrderSearchResult.cs
```diff
@@ -0,0 +1,17 @@
+using System.Collections.Generic;
+using VirtoCommerce.Storefront.Model.Order;
+
+namespace VirtoCommerce.Storefront.Model.Common.SearchResults
+{
+    public class CustomerOrderSearchResult
```
artem-dudarev:
```csharp
pulic class CustomerOrderSearchResult : GenericSearchResult<CustomerOrder>
{
}
```
The same for other search results

VirtoCommerce.Storefront.Model/Common/SearchResults/CustomerOrderSearchResult.cs
```diff
@@ -0,0 +1,17 @@
+using System.Collections.Generic;
+using VirtoCommerce.Storefront.Model.Order;
+
+namespace VirtoCommerce.Storefront.Model.Common.SearchResults
```
artem-dudarev: Don't put all search results in on folder

VirtoCommerce.Storefront/Controllers/Api/ApiAccountController.cs
```diff
@@ -280,7 +282,8 @@ public ActionResult<Organization> GetCustomerOrganization()
         [HttpPost("organization/users/search")]
         [Authorize(SecurityConstants.Permissions.CanViewUsers)]
         [ValidateAntiForgeryToken]
-        public async Task<ActionResult<GenericSearchResult<User>>> SearchOrganizationUsersAsync([FromBody] OrganizationContactsSearchCriteria searchCriteria)
+        [ResponseType(typeof(User))]
```
yecli: Probably should be UserSearchResult

PR #232, opened 2019-04-03
Add node installition record

kamilgarey:
Just add information about node installition


kamilgarey:
It should be done in dev branch

PR #241, opened 2019-04-16
removed Addresses field from user class

adminoren:
User object is converted to json with BackwardCompatibilityJsonConverter. It merges fields. As a result, the object wich contains one item in Addresses collection is converted to json with two items in Addresses.


tatarincev:
@adminoren This is  logic  by design to merge  Contact  and  User properties  for API methods that return User. This logic is left for backward compatibility with old themes. You might  disable this Json converter in your solution if this behavior is not desired. 
https://github.com/VirtoCommerce/vc-storefront-core/blob/master/VirtoCommerce.Storefront/JsonConverters/BackwardCompatibilityJsonConverter.cs#L83

PR #246, opened 2019-04-18
Users caching optimization

AliveMen:
refactoring PollingApiUserChangeToken
update PlatformAPI


VirtoCommerce.Storefront/Domain/Security/PollingApiUserChangeToken.cs
```diff
@@ -1,31 +1,36 @@
...
+        private PollingApiUserChangeToken(ISecurity platformSecurityApi, TimeSpan pollingInterval)
         {
-            _user = user;
-            _lastCheckedTimeUtc = DateTime.UtcNow;
             _pollingInterval = pollingInterval;
             _platformSecurityApi = platformSecurityApi;
         }
 
-        public static IChangeToken CreateChangeToken(User user, ISecurity platformSecurityApi, TimeSpan pollingInterval)
+        public static IChangeToken CreateChangeToken(ISecurity platformSecurityApi, TimeSpan pollingInterval)
```
yecli: Probably we could just use public constructor instead of static `CreateChangeToken` method.

PR #250, opened 2019-04-28
Fuzzy search parameter parsing from query string added to catalog search criterias

asvishnyakov:
Sorry, I forgot to include these changes into PR #120. Without them you can't use this feature in theme.

### DIFF
--- a/VirtoCommerce.Storefront.Model/Catalog/CategorySearchCriteria.cs
+++ b/VirtoCommerce.Storefront.Model/Catalog/CategorySearchCriteria.cs
@@ -42,6 +42,7 @@ namespace VirtoCommerce.Storefront.Model.Catalog
 
         private void Parse(NameValueCollection queryString)
         {
+            IsFuzzySearch = queryString.Get("fuzzy").EqualsInvariant(bool.FalseString);
             Keyword = queryString.Get("q");
             SortBy = queryString.Get("sort_by");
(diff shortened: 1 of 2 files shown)


VirtoCommerce.Storefront.Model/Catalog/CategorySearchCriteria.cs
```diff
@@ -42,6 +42,7 @@ public CategorySearchCriteria(Language language, NameValueCollection queryString
 
         private void Parse(NameValueCollection queryString)
         {
+            IsFuzzySearch = queryString.Get("fuzzy").EqualsInvariant(bool.FalseString);
```
yecli: Why comparing with False? Thus we set IsFuzzySearch=true only when passing fuzzy=false.

VirtoCommerce.Storefront.Model/Catalog/ProductSearchCriteria.cs
```diff
@@ -62,11 +62,10 @@ public override object Clone()
         }
 
         private void Parse(NameValueCollection queryString)
-        {
+        {        
+            IsFuzzySearch = queryString.Get("fuzzy").EqualsInvariant(bool.FalseString);
```
yecli: Why comparing with False? Thus we set IsFuzzySearch=true only when passing fuzzy=false.

VirtoCommerce.Storefront.Model/Catalog/CategorySearchCriteria.cs
asvishnyakov: Ooops, sorry, late night PR :)

PR #256, opened 2019-05-06
fix  CompatibilityVersion.Version_2_2 routing error

AliveMen:
do not use Endpoint Routing logic

### DIFF
--- a/VirtoCommerce.Storefront.Model/VirtoCommerce.Storefront.Model.csproj
+++ b/VirtoCommerce.Storefront.Model/VirtoCommerce.Storefront.Model.csproj
@@ -22,7 +22,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentValidation" Version="8.0.100" />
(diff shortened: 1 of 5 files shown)


VirtoCommerce.Storefront.Model/VirtoCommerce.Storefront.Model.csproj
```diff
@@ -22,7 +22,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentValidation" Version="8.0.100" />
-    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.2.1" />
+    <PackageReference Include="Microsoft.AspNetCore.App"/>
```
artem-dudarev: Don't remove space befoe `/>`

AliveMen: done

PR #257, opened 2019-05-07
store & language routing parameters marked as optional

asvishnyakov:
These optional routing parameters should use ASP.NET routing convention (i.e. `/{store?}/{language?}`) and be marked as optional in swagger


asvishnyakov:
Closed as violate OpenAPI specification

> If the parameter is in "path", this property is required and its value MUST be true.

PR #273, opened 2019-07-01
date filter fix

trueboroda:


VirtoCommerce.Storefront.Model/Order/Services/ICustomerOrderService.cs
```diff
@@ -11,5 +11,6 @@ public interface ICustomerOrderService
         Task<IPagedList<CustomerOrder>> SearchOrdersAsync(OrderSearchCriteria criteria);
         IPagedList<CustomerOrder> SearchOrders(OrderSearchCriteria criteria);
         Task<CustomerOrder> GetOrderByNumberAsync(string number);
+		Task<CustomerOrder> GetOrderByIdAsync(string number);
```
artem-dudarev: Don't use editor that doesn't support `.editorconfig`

VirtoCommerce.Storefront/Domain/Marketing/MarketingService.cs
```diff
@@ -1,69 +1,47 @@
...
         }
 
+        [Obsolete("Use IDynamicContentEvaluator instead")]
         public virtual async Task<string> GetDynamicContentHtmlAsync(string storeId, string placeholderName)
         {
             string htmlContent = null;
 
-            //TODO: make full context
-            var evaluationContext = new DynamicContentEvaluationContext
+            var evalContext = _workContextAccessor.WorkContext.ToDynamicContentEvaluationContext();
```
artem-dudarev: Don't cut words. Should be `evaluationContext `

VirtoCommerce.Storefront/Routing/SlugRouteService.cs
```diff
@@ -201,10 +201,7 @@ protected virtual async Task LoadObjectsAndBuildFullSeoPaths(string objectType,
             {
                 foreach (var seo in objects)
                 {
-                    if (seoPaths.ContainsKey(seo.ObjectId))
-                    {
-                        seo.SeoPath = seoPaths[seo.ObjectId];
-                    }
+                    seo.SeoPath = seoPaths[seo.ObjectId] ?? seo.SeoPath;
```
artem-dudarev: This is bad practice. If you don't add `.WithDefaultValue(null)` you'll get an exception.
Create an extension method wich will check if the key exists and use it like this:
`seoPaths.GetValueSafe(seo.ObjectId, null) ?? seo.SeoPath;`
``` csharp
    public static class DictionaryExtensions
    {
        public static T GetValueSafe<T>(this IDictionary<string, T> dictionary, string key, T defaultValue)
        {
            T result;

            if (key != null && dictionary.TryGetValue(key, out var value))
            {
                result = value;
            }
            else
            {
                result = defaultValue;
            }

            return result;
        }

        public static bool TryGetValueSafe<T>(this IDictionary<string, T> dictionary, string key, out T value)
        {
            bool result;

            if (key != null)
            {
                result = dictionary.TryGetValue(key, out value);
            }
            else
            {
                result = false;
                value = default(T);
            }

            return result;
        }
    }
```

PR #286, opened 2019-08-02
Added  tax details support

adminoren:


VirtoCommerce.Storefront.Model/Cart/LineItem.cs
```diff
@@ -292,13 +292,23 @@ public void ApplyTaxRates(IEnumerable<TaxRate> taxRates)
...
+                {
+                    var amount = ExtendedPrice.Amount > 0 ? ExtendedPrice.Amount : SalePrice.Amount;
+                    if (amount > 0)
+                    {
+                        TaxPercentRate = TaxRate.TaxPercentRound(lineItemTaxRate.Rate.Amount / amount);
+                    }
+                }
+
+                TaxDetails.Clear();
+                TaxDetails.AddRange(lineItemTaxRate.Line.TaxDetails);
```
tatarincev: Can you replace this to the simple assignment statement? `TaxDetails = `

VirtoCommerce.Storefront.Model/Tax/TaxDetail.cs
```diff
@@ -6,10 +6,9 @@ public partial class TaxDetail : ValueObject
     {
         public TaxDetail(Currency currency)
         {
-            Rate = new Money(currency);
             Amount = new Money(currency);
         }
-        public Money Rate { get; set; }
+        public decimal Rate { get; set; }
```
tatarincev: Why do you need this type change?

VirtoCommerce.Storefront.Model/Tax/TaxLine.cs
```diff
@@ -30,5 +27,7 @@ public TaxLine(Currency currency)
         /// </summary>
         public Money Price { get; set; }
         public string TaxType { get; set; }
+        public string TypeName { get; set; }
+        public ICollection<TaxDetail> TaxDetails { get; set; }
```
tatarincev: better to use property assignment here  ` = new List`

VirtoCommerce.Storefront.Model/Tax/TaxRate.cs
```diff
@@ -10,6 +10,7 @@ public TaxRate(Currency currency)
             Rate = new Money(currency);
         }
         public Money Rate { get; set; }
+        public decimal PercentRate { get; set; }
```
tatarincev: Why do you need this type change?

VirtoCommerce.Storefront.Model/Tax/TaxLine.cs
artem-dudarev: And change `ICollection` to `IList`.
Always use IList - it is shorter and allows to access items by index.

VirtoCommerce.Storefront.Model/Tax/TaxRate.cs
adminoren: PercentRate is not money, it's a number. Database field also has decimal instead of money, so I decided to change it

VirtoCommerce.Storefront.Model/Tax/TaxLine.cs
adminoren: ok, I' fix it

PR #295, opened 2019-08-30
User can get any order

trueboroda:
Customer user can get any order if has next permission "storefront:order:view".

### DIFF
--- a/VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
+++ b/VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
@@ -190,13 +190,15 @@ namespace VirtoCommerce.Storefront.Controllers.Api
         private async Task<orderModel.CustomerOrder> GetOrderDtoByNumber(string number)
         {
             var order = await _orderApi.GetByNumberAsync(number);
-            var authorizationResult = await _authorizationService.AuthorizeAsync(User, order, CanAccessOrderAuthorizationRequirement.PolicyName);
-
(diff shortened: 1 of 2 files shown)


VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
```diff
@@ -190,13 +190,15 @@ private async Task<CustomerOrder> GetOrderByNumber(string number)
...
         {
             var order = await _orderApi.GetByNumberAsync(number);
-            var authorizationResult = await _authorizationService.AuthorizeAsync(User, order, CanAccessOrderAuthorizationRequirement.PolicyName);
-
-            if (!authorizationResult.Succeeded)
+            var viewAllOrdersAuthorizationResult = await _authorizationService.AuthorizeAsync(User, null, SecurityConstants.Permissions.CanViewOrders);
+            if (!viewAllOrdersAuthorizationResult.Succeeded)
             {
-                throw new StorefrontException($"Order with number {{ number }} not found (or not belongs to current user)");
+                var authorizationResult = await _authorizationService.AuthorizeAsync(User, order, CanAccessOrderAuthorizationRequirement.PolicyName);
```
tatarincev: Why you need this double authorization check? Why it can't be done  by one policy `CanAccessOrderAuthorizationRequirement`?

trueboroda: Ok. I will think about this.

PR #296, opened 2019-09-02
Order status changing for the support of PP workflow

t13ka:
This method is needed to support of procurement portal workflow in the case when we try to change only the order status. Predefined order statuses aren't provided.


VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
```diff
@@ -180,6 +180,28 @@ public async Task<ActionResult> GetInvoicePdf(string orderNumber)
             return File(stream, "application/pdf");
         }
 
+        // PUT: storefrontapi/orders/{orderNumber}/status
+        [HttpPut("{orderNumber}/status")]
+        [ValidateAntiForgeryToken]
+        public async Task<ActionResult> ChangeOrderStatus(string orderNumber, [FromBody] dynamic payload)
+        {
+            var newStatus = payload.newStatus.ToString();
```
artem-dudarev: How do you know that `dynamic payload` has `newStatus` field?

VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
```diff
@@ -180,6 +180,28 @@ public async Task<ActionResult> GetInvoicePdf(string orderNumber)
...
+
+            if (string.IsNullOrEmpty(newStatus))
+            {
+                throw new StorefrontException("The new order status can't be nullable or empty");
+            }
+
+            using (await AsyncLock.GetLockByKey(GetAsyncLockKey(orderNumber, WorkContext)).LockAsync())
+            {
+                var orderDto = await GetOrderDtoByNumber(orderNumber);
+                orderDto.Status = newStatus;
```
artem-dudarev: Any user can set any status?

VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
tatarincev: Definitely you should to avoid of using the `dynamic` type  as type for formal  parameters for API methods,  since this type doesn't have a definition in  the `OpenApi`  standard

VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
tatarincev: You must add authorization for this API, it will be the 100% vulnerability if any user with any privileges (even anonymous) can change the status of any order in the system simply by calling this API.

artem-dudarev: `GetOrderDtoByNumber()` checks if current user can access the order, but even if this is true it doesn't mean current user can modify this order.

VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
t13ka: Actually, I've added dynamic type here only to simplify to limit the diffs and to localize all changes in one method and do not add extra classes and so on. If you think that adding extra request models it's ok in this case then we certainly will add them)
Also, we know for sure that payload will contain this field because we build the payload on the client-side.

VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
t13ka: The idea is that user management layer (for procurement portal) will be built on the client-side and we going to permit order status changing only for platform users with appropriate rights (must be admin). In this case, the admin credential will be as a proxy. How we should verify our permissions in the right way to handle this case?

VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
```diff
@@ -180,6 +180,33 @@ public async Task<ActionResult> GetInvoicePdf(string orderNumber)
...
+        public async Task<ActionResult> ChangeOrderStatus(string orderNumber, [FromBody] OrderStatusChange payload)
+        {
+            if (!ModelState.IsValid)
+            {
+                return BadRequest(ModelState);
+            }
+
+            using (await AsyncLock.GetLockByKey(GetAsyncLockKey(orderNumber, WorkContext)).LockAsync())
+            {
+                var order = await _orderApi.GetByNumberAsync(orderNumber);
```
tatarincev: You are loading order twice. Is it really necessary?  I can't see any obstacles to check authorization  for order loaded by call `GetOrderDtoByNumber` method

VirtoCommerce.Storefront/Controllers/Api/ApiOrderController.cs
```diff
@@ -180,6 +180,33 @@ public async Task<ActionResult> GetInvoicePdf(string orderNumber)
             return File(stream, "application/pdf");
         }
 
+        // PUT: storefrontapi/orders/{orderNumber}/status
+        [HttpPut("{orderNumber}/status")]
+        [ValidateAntiForgeryToken]
+        public async Task<ActionResult> ChangeOrderStatus(string orderNumber, [FromBody] OrderStatusChange payload)
```
tatarincev: Please rename payload formal parameter  to something else to be more concrete

VirtoCommerce.Storefront/Domain/Security/OrderStatusChangeHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+    using System.Linq;
+    using System.Threading.Tasks;
+
+    using Microsoft.AspNetCore.Authorization;
+
+    using VirtoCommerce.Storefront.AutoRestClients.OrdersModuleApi.Models;
+    using VirtoCommerce.Storefront.Model;
+    using VirtoCommerce.Storefront.Model.Security;
+
+    public class OrderStatusChangeHandler : AuthorizationHandler<OrderStatusChangeAuthorizationRequirement, CustomerOrder>
```
tatarincev: I  suppose that we need a one common `AuthorizationHandler`  for  check  authorization based on the permissions policies.  That can be used as method attribute `[Authorization('permission-name')` 

See how this has done in the platform v3:

https://github.com/VirtoCommerce/vc-platform-core/blob/dev/VirtoCommerce.Platform.Web/Controllers/Api/ModulesController.cs#L58
https://github.com/VirtoCommerce/vc-platform-core/blob/dev/VirtoCommerce.Platform.Security/Authorization/PermissionAuthorizationPolicyProvider.cs
https://github.com/VirtoCommerce/vc-platform-core/blob/dev/VirtoCommerce.Platform.Security/Authorization/PermissionAuthorizationHandlerBase.cs

Issue #314, opened 2019-09-30
'Sort by' combo in product list does nothing


akak1977:
![image](https://user-images.githubusercontent.com/9972421/65862876-923c6600-e388-11e9-8b7f-824eced3f19c.png)

yecli:
https://github.com/VirtoCommerce/vc-storefront-core/releases/tag/v4.1.2

PR #323, opened 2019-10-23
fix sortBy prop assigning for productSearchcriteria

ilyawzrd:
'Sort by' feature didn't work because sortBy is constantly being replaced by same value
fix for #314

### DIFF
--- a/VirtoCommerce.Storefront/Domain/Catalog/CatalogWorkContextBuilderExtensions.cs
+++ b/VirtoCommerce.Storefront/Domain/Catalog/CatalogWorkContextBuilderExtensions.cs
@@ -14,13 +14,17 @@ namespace VirtoCommerce.Storefront.Domain
             var serviceProvider = builder.HttpContext.RequestServices;
             var catalogService = serviceProvider.GetRequiredService<ICatalogService>();
             var workContext = builder.WorkContext;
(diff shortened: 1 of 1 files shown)


VirtoCommerce.Storefront/Domain/Catalog/CatalogWorkContextBuilderExtensions.cs
```diff
@@ -18,9 +18,9 @@ public static Task WithCatalogsAsync(this IWorkContextBuilder builder)
             //Initialize catalog search criteria
             var productSearchcriteria = new ProductSearchCriteria(workContext.CurrentLanguage, workContext.CurrentCurrency, workContext.QueryString)
             {
-                UserGroups = workContext.CurrentUser?.Contact?.UserGroups ?? new List<string>(),
-                SortBy = "priority-descending;name-ascending"
+                UserGroups = workContext.CurrentUser?.Contact?.UserGroups ?? new List<string>()                               
             };
+            productSearchcriteria.SortBy = productSearchcriteria.SortBy ?? "priority-descending;name-ascending";
```
yecli: I think better to check using string.IsNullOrEmpty(), as "" value should be treat as default sorting.

PR #328, opened 2019-10-30
BE: Fix for the coupons that have conditions for parent category.

blazemachine:


VirtoCommerce.Storefront/Domain/Cart/CartConverter.cs
```diff
@@ -407,7 +407,8 @@ public static PromotionEvaluationContext ToPromotionEvaluationContext(this Shopp
                 User = cart.Customer,
                 Currency = cart.Currency,
                 Language = cart.Language,
-                StoreId = cart.StoreId
+                StoreId = cart.StoreId,
```
tatarincev: Seems these lines are redundant because of this line do this for line items  https://github.com/VirtoCommerce/vc-storefront-core/blob/master/VirtoCommerce.Storefront/Domain/Marketing/MarketingConverter.cs#L147

And you need to add the missing set of the `Outline` property from the` LineItem.Product` in this converter https://github.com/VirtoCommerce/vc-storefront-core/blob/master/VirtoCommerce.Storefront/Domain/Cart/CartConverter.cs#L787

PR #353, opened 2019-11-21
disllow log in under the other store account

trueboroda:


yecli:
We cannot log in to the other store now.
But the problem in issue tells that we should sign out when changing the store, so the fix does not solve the described problem.

yecli:
Release https://github.com/VirtoCommerce/vc-storefront-core/releases/tag/v4.1.4

PR #354, opened 2019-11-25
vp-1015 fix customer list name

AliveMen:


artem-dudarev:
This fix is incorrect.
You should unescape list name in `ApiListsController.cs` in every method that has `{listName}` in the route, because ASP.NET Core doesn't unescape route paramters automatically.
https://github.com/aspnet/Mvc/issues/4599

yecli:
If we fix it that way, it will only fix new list names. Probably we need to show unescaped name for existing lists with a space in the name too. If it is an important case, I think we need to handle unescaped names in converting to the storefront model.

yecli:
After internal discussion with PO decided that existing lists should not be fixed here, we need to make proper saving for the new ones.
So will fix as proposed by @artem-dudarev.

PR #363, opened 2019-12-23
fixes for scriban

basilkot:


basilkot:
@t13ka 

> one question: why you rename the method 'Shortcut' in ArrayFilters.cs to 'SortList'?

Scriban has internal `sort` filter that cannot be overriden.

PR #366, opened 2019-12-30
Fix memory leak and cache corruption

artem-dudarev:


yecli:
Released in https://github.com/VirtoCommerce/vc-storefront-core/releases/tag/v4.1.6

PR #372, opened 2020-01-22
vp-1260_loquid_collection_product_properties

AliveMen:
Fill catalog.products.properties when filling related products


yecli:
Not needed, could be resolved in liquid template by settings response group for `paginator`

PR #375, opened 2020-01-28
vp-1079 fix Anti Forgery Token issue when reload login form.

AliveMen:
LogIn
Refactoring Login method.

### DIFF
--- a/VirtoCommerce.Storefront/Controllers/AccountController.cs
+++ b/VirtoCommerce.Storefront/Controllers/AccountController.cs
@@ -312,7 +312,6 @@ namespace VirtoCommerce.Storefront.Controllers
             if (!new CanUserLoginToStoreSpecification(user).IsSatisfiedBy(WorkContext.CurrentStore) || new IsUserSuspendedSpecification().IsSatisfiedBy(user))
             {
                 WorkContext.Form.Errors.Add(SecurityErrorDescriber.UserCannotLoginInStore());
-                WorkContext.Form.Errors.Add(SecurityErrorDescriber.LoginFailed());
                 return View("customers/login", WorkContext);
             }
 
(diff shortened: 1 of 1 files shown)


VirtoCommerce.Storefront/Controllers/AccountController.cs
```diff
@@ -301,101 +301,100 @@ public async Task<ActionResult> Login([FromForm] Login login, [FromQuery]string
...
-            if (loginResult.IsLockedOut)
-            {
-                return View("lockedout", WorkContext);
-            }
-
-            if (loginResult is CustomSignInResult signInResult && signInResult.IsRejected)
+            else
             {
-                WorkContext.Form.Errors.Add(SecurityErrorDescriber.AccountIsBlocked());
+                WorkContext.Form.Errors.Add(SecurityErrorDescriber.OperationFailed());
```
yecli: Since we do not follow the single return rule in this controller, to improve readability please make handling of 
```
if (user == null) 
{
…
}
```
right after getting the user.

yecli: Let's return `SecurityErrorDescriber.LoginFailed()` instead of `SecurityErrorDescriber.OperationFailed()` in case of non-existent user, to unify wrong password result woth non existent user.

AliveMen: Done

VirtoCommerce.Storefront/Controllers/AccountController.cs
```diff
@@ -301,101 +301,100 @@ public async Task<ActionResult> Login([FromForm] Login login, [FromQuery]string
...
-            var loginResult = await _signInManager.PasswordSignInAsync(login.UserName, login.Password, login.RememberMe, lockoutOnFailure: true);
+            var user = await _signInManager.UserManager.FindByNameAsync(login.UserName);
 
-            if (loginResult.Succeeded)
+            if (user != null)
             {
-                var user = await _signInManager.UserManager.FindByNameAsync(login.UserName);
-
-                // Check that current user can sing in to current store
                 if (new CanUserLoginToStoreSpecification(user).IsSatisfiedBy(WorkContext.CurrentStore) && new IsUserSuspendedSpecification().IsSatisfiedBy(user) == false)
```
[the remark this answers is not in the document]
AliveMen: fixed

PR #377, opened 2020-01-31
chore #CCIT-301: add troubleshooting info into !readme.txt 

YanBrovkin:
Sometimes, during generating autorest client of your module using Swagger, the error 'Could not read 'http://...' could appear.
To prevent this error you have to generate Swagger description as a json file at first. 
And then use it as -input parameter for generating autorest client


YanBrovkin:
> 
> 
> wrong branch, should be dev

switched to dev

PR #379, opened 2020-02-12
Added source maps support

asvishnyakov:


yecli:
Seems no longer actual

PR #382, opened 2020-02-28
VP-1527: Make proper redirect links when store has subfolder in url

yecli:
- There was a problem when store has subfolder in the url (e.g. for store url "http://localhost/store" - subfolder "store"). Redirect links were constructed incorrectly. Now store subfolder is taken into account.


yecli:
Released https://github.com/VirtoCommerce/vc-storefront-core/releases/tag/v4.1.8

PR #383, opened 2020-02-28
VP-1558: Merge settings between base and current themes

asvishnyakov:
Fully backward-compatible: see changes in settings
Support merge between themes in different stores and themes in the same store


yecli:
https://github.com/VirtoCommerce/vc-storefront-core/releases/tag/v4.1.8

PR #386, opened 2020-03-05
Vp 1629 character escaping on search bar

AliveMen:
Remove query escaping from StoreFront


yecli:
https://github.com/VirtoCommerce/vc-storefront-core/releases/tag/v4.1.8

PR #389, opened 2020-03-16
Look for presets in base theme if current preset specified in current theme

asvishnyakov:
image


VirtoCommerce.LiquidThemeEngine/ShopifyLiquidThemeEngine.cs
```diff
@@ -358,13 +358,26 @@ public ValueTask<string> RenderTemplateAsync(string templateContent, string temp
 
                 JObject result;
                 var baseThemeSettings = new JObject();
-                var currentThemeSettings = result = GetCurrentSettingsPreset(InnerGetAllSettings(_themeBlobProvider, CurrentThemeSettingPath));
+                var allCurrentThemeSettings = InnerGetAllSettings(_themeBlobProvider, CurrentThemeSettingPath);
+                var currentThemeSettings = result = null;
+                try
+                {
+                    currentThemeSettings = result = GetCurrentSettingsPreset(allCurrentThemeSettings);
```
basilkot: The currentThemeSettings will be null when child settings file hasn't presets. It will cause the problem. see comment below.

basilkot: The currentThemeSettings will be null when child settings file hasn't presets. It will cause the problem on row 386. The merge process will be occurred between base theme and this null object.

asvishnyakov: Fixed, added test

PR #392, opened 2020-03-19
fix checking settings file prefix to empty value

basilkot:


VirtoCommerce.LiquidThemeEngine/ShopifyLiquidThemeEngine.cs
```diff
@@ -497,7 +497,7 @@ private string ReadTemplateByPath(string templatePath)
         private string GetSettingsFilePath()
         {
             var prefix = _httpContextAccessor.HttpContext.Request.Query["preview_mode"];
-            return prefix.IsNullOrEmpty() ? "settings_data.json" : $"drafts\\{prefix}_settings_data.json";
+            return prefix.ToString().IsNullOrEmpty() ? "settings_data.json" : $"drafts\\{prefix}_settings_data.json";
```
tatarincev: potential null reference exception

basilkot: No, it always has value

[QueryCollection.Item](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.querycollection.item?view=aspnetcore-3.1)

PR #394, opened 2020-03-23
VP-1257: Wrong value in country field

AliveMen:
Fix LINQ query condition.


VirtoCommerce.Storefront/Domain/Countries/FileSystemCountriesService.cs
```diff
@@ -1,22 +1,21 @@
...
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
-using VirtoCommerce.Storefront.Caching;
+using Microsoft.Extensions.Caching.Memory;
```
yecli: Seems some of new using are redundant

PR #402, opened 2020-04-09
Add missing methods for working with lists

ilyawzrd:
* ChangeListItem for changing quantity
* ClearList for removing all items from list

### DIFF
--- a/VirtoCommerce.Storefront/Controllers/Api/ApiListsController.cs
+++ b/VirtoCommerce.Storefront/Controllers/Api/ApiListsController.cs
@@ -99,10 +99,10 @@ namespace VirtoCommerce.Storefront.Controllers.Api
             }
         }
 
-        // PUT: storefrontapi/lists/{listName}/{type}/items/{lineItemId}?quantity=...
-        [HttpPut("{listName}/{type}/items/{lineItemId}")]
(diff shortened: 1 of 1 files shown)


VirtoCommerce.Storefront/Controllers/Api/ApiListsController.cs
```diff
@@ -99,6 +99,43 @@ public async Task<ActionResult<ShoppingCartItems>> RemoveItemFromList(string lin
             }
         }
 
+        // PUT: storefrontapi/lists/{listName}/{type}/items/{lineItemId}?quantity=...
+        [HttpPut("{listName}/{type}/items/{lineItemId}")]
+        [ValidateAntiForgeryToken]
+        public async Task<ActionResult> ChangeListItem(string listName, string type, string lineItemId, int quantity)
```
tatarincev, line 105: Instead of pass lineItemId and quantity as parameters better use https://github.com/VirtoCommerce/vc-storefront-core/blob/dev/VirtoCommerce.Storefront.Model/Cart/ChangeCartItemQty.cs type. and don't forget to add binding attributes to al parameters [FromRoute] and [FromBody]

yecli:
Released: https://github.com/VirtoCommerce/vc-storefront-core/releases/tag/v4.1.11

PR #406, opened 2020-04-22
Initial implementation of the toggling features process

t13ka:
Now we have a features agent which reading file wwwroot/cms-content/settings_data.json
This agent reads the file and evaluates the results which can be rendered on liquid pages


VirtoCommerce.Storefront.Tests/Features/Samples/full_data.json
```diff
@@ -0,0 +1,156 @@
+{
+  "features": [
+    {
+      "Name": "OrdersBrowsing",
+      "IsActive": true,
+      "Requires": [],
+      "Replaces": [],
+      "Replaced": [],
```
trueboroda, line 8: Why are these two parameters necessary? Is this not redundant? Two sources of truth.

t13ka: Agree. The replaced property was removed

VirtoCommerce.LiquidThemeEngine/Filters/FeatureFilter.cs
```diff
@@ -0,0 +1,65 @@
+namespace VirtoCommerce.LiquidThemeEngine.Filters
+{
+    using System;
```
tatarincev, line 3: Could you do not use this nested `using` declaration? This syntax looks to be inconsistent with our general code style convention. We have negotiated regarding this style in one of dev meetings and have come to an agreement to do not use nested `using` statements.

VirtoCommerce.LiquidThemeEngine/Filters/FeatureFilter.cs
```diff
@@ -0,0 +1,65 @@
...
+    public class FeatureFilter
+    {
+        public static string IsFeatureActive(TemplateContext context, string key, params object[] variables)
+        {
+            if (string.IsNullOrEmpty(key))
+            {
+                return IsFeaturesActive(context, key, variables);
+            }
+
+            string result;
```
tatarincev, line 18: this condition may be simplified by default initialization
```C#
var result = Convert.ToString(false);
if (context.TemplateLoader is ShopifyLiquidThemeEngine themeEngine)
{
...
}
``

File: VirtoCommerce.LiquidThemeEngine/Filters/FeatureFilter.cs [hunk 4]
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        public static string IsFeaturesActive(TemplateContext context, string key, params object[] variables)
+        {
+            if (!string.IsNullOrEmpty(key))
+            {
+                return IsFeatureActive(context, key, variables);
+            }
+
+            string result;
```
tatarincev, line 38: see above

VirtoCommerce.LiquidThemeEngine/Filters/FeatureFilter.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    using System;
+
+    using global::Scriban;
+
+    using Newtonsoft.Json.Linq;
+
+    public class FeatureFilter
+    {
+        public static string IsFeatureActive(TemplateContext context, string key, params object[] variables)
```
tatarincev, line 11: You can return any Built-in .NET types from Liquid filters, you do not need to return only string then.

VirtoCommerce.LiquidThemeEngine/ShopifyLiquidThemeEngine.cs
```diff
@@ -499,5 +502,24 @@ private string GetSettingsFilePath()
             var prefix = _httpContextAccessor.HttpContext.Request.Query["preview_mode"];
             return prefix.ToString().IsNullOrEmpty() ? "settings_data.json" : $"drafts\\{prefix}_settings_data.json";
         }
+
+        public bool IsFeatureActive(string featureName)
+        {
+            return _memoryCache.GetOrCreateExclusive(featureName, cacheEntry =>
```
tatarincev, line 508: Better create cache key by this statement. This is guarantee its uniqueness
```
 var cacheKey = CacheKey.With(GetType(), nameof(IsFeatureActive), featureName);
```

VirtoCommerce.LiquidThemeEngine/Filters/FeatureFilter.cs
t13ka: done

VirtoCommerce.LiquidThemeEngine/Filters/FeatureFilter.cs
t13ka: I simplified the entire code

VirtoCommerce.LiquidThemeEngine/Filters/FeatureFilter.cs
t13ka: see above

VirtoCommerce.LiquidThemeEngine/ShopifyLiquidThemeEngine.cs
t13ka: done

PR #407, opened 2020-04-24
Feature/vp 1417 upgrade to core 3 1

akak1977:
- Upgrade to core 3.1;
- Upgrade restful proxies (client APIs) to latest platform/modules v3 endpoints;
- Codestyle changes.


VirtoCommerce.Storefront/Domain/Common/SettingConverter.cs
```diff
@@ -1,22 +1,23 @@
...
             var retVal = new SettingEntry();
-            retVal.DefaultValue = settingDto.DefaultValue;
-            retVal.Description = settingDto.Description;
-            retVal.IsArray = settingDto.IsArray ?? false;
+            retVal.DefaultValue = settingDto.DefaultValue?.ToString();
+            retVal.IsArray = false;
             retVal.Name = settingDto.Name;
-            retVal.Title = settingDto.Title;
-            retVal.Value = settingDto.Value;
+            retVal.Value = settingDto.Value?.ToString();
```
tatarincev, line 15: You can't use Value.ToString() to cast object value. The ToString method of any object is supposed to return a string representation of that object and can produces  a different object string representation based on execution context thread culture for example. e.g  3.15.ToString() => "3,14" and after back conversation to decimal with invariant culture  will produce 314 decimal value. What exactly happens for tax settings.  https://github.com/VirtoCommerce/vc-storefront-core/blob/master/VirtoCommerce.Storefront/Domain/Stores/StoreService.cs#L118

tatarincev: I propose to change the type to from `string` to `object` for these properties  `SettingEntry.Value`, `SettingEntry.DefaultValue`, `SettingEntry.AllowedValues` to prevent the object to string casting errors.

VirtoCommerce.Storefront/Domain/Security/UserStoreStub.cs
```diff
@@ -435,25 +451,6 @@ public Task<IList<User>> GetUsersForClaimAsync(Claim claim, CancellationToken ca
 
         #region IRoleStore<Role> members
 
-        public async Task<IdentityResult> CreateAsync(Role role, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
akak1977: It's something SLint-related.

PR #408, opened 2020-04-24
Logic for prepare content pages and articles

basilkot:
These changes allow to display blog articles that was created in page builder


VirtoCommerce.Storefront/Domain/StaticContent/StaticContentItemBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+        public StaticContentItemBuilder(IStaticContentItemFactory factory)
+        {
+            _factory = factory;
+        }
+
+        public ContentItem BuildFrom(string baseStoreContentPath, string blobRelativePath, string content)
+        {
+            var contentItem = _factory.GetItemFromPath(blobRelativePath);
+            var visitedContent = content;
+            Action<List<IContentItemVisitor>> runParsers = parsers =>
```
tatarincev, line 52: rename to visitors

VirtoCommerce.Storefront/Domain/StaticContent/StaticContentItemBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace VirtoCommerce.Storefront.Domain
+{
+    public partial class StaticContentItemBuilder : IStaticContentItemBuilder
+    {
+        internal static readonly Regex headerRegExp = new Regex(@"(?s:^---(.*?)---)");
+        internal static readonly string[] extensions = new[] { ".md", ".liquid", ".html" };
+
+        private readonly IStaticContentItemFactory _factory;
+
+        private readonly List<IContentItemVisitor> _prepareVisitors = new List<IContentItemVisitor>
```
tatarincev, line 19: I think It would be better to inject all visitors as array of dependencies in the ctor. Why do you need to group them if anyway you call all visitors in the one loop?

VirtoCommerce.Storefront/Domain/StaticContent/StaticContentItemBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+        public ContentItem BuildFrom(string baseStoreContentPath, string blobRelativePath, string content)
+        {
+            var contentItem = _factory.GetItemFromPath(blobRelativePath);
+            var visitedContent = content;
+            Action<List<IContentItemVisitor>> runParsers = parsers =>
+            {
+                parsers.Where(x => x.Suit(contentItem))
+                    .ForEach(x => visitedContent = x.ReadContent(blobRelativePath, visitedContent, contentItem));
+            };
+            if (contentItem != null)
```
tatarincev, line 57: I think this "if" the check is in the not proper place, it must be above,  before running visitors in line 51

VirtoCommerce.Storefront/Domain/StaticContent/StaticContentItemBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                contentItem.StoragePath = "/" +
+                    (string.IsNullOrEmpty(baseStoreContentPath) ?
+                        blobRelativePath :
+                        blobRelativePath.Replace(baseStoreContentPath + "/", String.Empty)
+                    ).TrimStart('/');
+                contentItem.FileName = Path.GetFileName(blobRelativePath);
+                new List<List<IContentItemVisitor>>
```
tatarincev, line 65: why do we need to run all visitors again? Is it only to take into account the storage patch changes that are introduced in line 59? I think it is can be optimized to do only by single bypass

VirtoCommerce.Storefront/Domain/StaticContent/StaticContentService.cs
```diff
@@ -83,82 +65,81 @@ public IEnumerable<ContentItem> LoadStoreStaticContent(Store store)
...
-                    }
-                }
-            }
+        #endregion
 
-            public string Name { get; }
-            public Language Language { get; }
-            public string Path { get; }
-        }
+        //private void LoadAndRenderContentItem(string contentPath, ContentItem contentItem)
```
tatarincev, line 82: remove these commented out lines

VirtoCommerce.Storefront/Domain/StaticContent/StaticContentItemBuilder.cs
basilkot: it is a variable, not processing. processing is executed on line 67

VirtoCommerce.Storefront/Domain/StaticContent/StaticContentItemBuilder.cs
basilkot: it is a single now. on line 52 it is a functional variable to run visitors

VirtoCommerce.Storefront/Domain/StaticContent/StaticContentItemBuilder.cs
basilkot: the visitors order has matter. there is no guarantees to save the order during injecting.
Will it be normal to add priority for visitors?

PR #409, opened 2020-04-27
VP-2067 Add login option to redirect user to correct store instead of returning error

asvishnyakov:


VirtoCommerce.Storefront/Controllers/AccountController.cs
```diff
@@ -309,7 +311,23 @@ public async Task<ActionResult> Login([FromForm] Login login, [FromQuery]string
                 return View("customers/login", WorkContext);
             }
 
-            if (!new CanUserLoginToStoreSpecification(user).IsSatisfiedBy(WorkContext.CurrentStore) || new IsUserSuspendedSpecification().IsSatisfiedBy(user))
+            if (!new CanUserLoginToStoreSpecification(user).IsSatisfiedBy(WorkContext.CurrentStore))
```
tatarincev, line 314: Why we can't do this by passing a returnUrl in the query string in form post action?
e.g `{{ '/' | store_absolute_url }}`

asvishnyakov: @tatarincev Because code will not pass to this action, it will not allow you to login to this store

asvishnyakov:
@tatarincev Refactored

New code is refactored version of already used code written by @yecli 

Also, I found you, @tatarincev introduced few breaking changes in https://github.com/VirtoCommerce/vc-storefront-core/commit/c436160807c6a07d4ce2e01ae95b1ad28fdb1654#diff-4572e31bd7dfce2f9afdc19c23dd0881, especially in https://github.com/VirtoCommerce/vc-storefront-core/commit/c436160807c6a07d4ce2e01ae95b1ad28fdb1654#diff-4572e31bd7dfce2f9afdc19c23dd0881 and https://github.com/VirtoCommerce/vc-storefront-core/commit/c436160807c6a07d4ce2e01ae95b1ad28fdb1654#diff-3ba295d1c2f3459ba047c83b0521b5d0

For example:
1. Your "fix" for store url breaks StoreUrlBuilder ToAppAbsolute and ToAppRelative methods work, because ToAppRelative will never generate urls like "~/...". but instead always produce "/..."
2. You just removed existed check for user suspension and user store check. I didn't found information or issue number in comment of https://github.com/VirtoCommerce/vc-storefront-core/commit/c436160807c6a07d4ce2e01ae95b1ad28fdb1654

Please pay attention and feel free to change this PR if needed

VirtoCommerce.Storefront/Domain/Security/SecurityWorkContextBuilderExtensions.cs
```diff
@@ -26,7 +27,19 @@ public static async Task WithCurrentUserAsync(this IWorkContextBuilder builder)
             }).ToList();
 
             var user = await signInManager.UserManager.GetUserAsync(builder.HttpContext.User);
-           
+
+            if (user != null && !new CanUserLoginToStoreSpecification(user).IsSatisfiedBy(builder.WorkContext.CurrentStore))
```
tatarincev, line 31: I've removed these lines by this commit https://github.com/VirtoCommerce/vc-storefront-core/commit/c436160807c6a07d4ce2e01ae95b1ad28fdb1654#diff-4572e31bd7dfce2f9afdc19c23dd0881L30  but unfortunately, I didn't write a unit-tests for this bug, but  I remember that this was due to problems with automatic user logout when using several stores. I would like to recommend removing these lines until we won't ensure that all works as expected.

VirtoCommerce.Storefront/Domain/Stores/StoreWorkContextBuilderExtensions.cs
```diff
@@ -21,13 +21,7 @@ public static Task WithStoresAsync(this IWorkContextBuilder builder, IEnumerable
             }
 
             builder.WorkContext.AllStores = stores.ToArray();
-            var currentStore = builder.HttpContext.GetCurrentStore(stores, defaultStoreId);
```
tatarincev, line 24: Could you explain why did you remove these lines of code?

VirtoCommerce.Storefront/Domain/Security/SecurityWorkContextBuilderExtensions.cs
asvishnyakov: @tatarincev We may replace it with Forbidden response but we need to prevent customer login into another store anyway.

VirtoCommerce.Storefront/Domain/Stores/StoreWorkContextBuilderExtensions.cs
asvishnyakov: Please read again, I already explained it:

> Your "fix" for store url breaks StoreUrlBuilder ToAppAbsolute and ToAppRelative methods, because ToAppRelative will never generate urls like ~/.... but instead always produce /...

VirtoCommerce.Storefront/Domain/Security/SecurityWorkContextBuilderExtensions.cs
asvishnyakov: Or I think it's better to remove sign out and leave user as anonymous

VirtoCommerce.Storefront/Domain/Stores/StoreWorkContextBuilderExtensions.cs
asvishnyakov: Reverted back my changes, will let you know if something broken

PR #412, opened 2020-04-30
#VP-1870 condition apply only full price

kostyrin:


kostyrin:
move to https://github.com/VirtoCommerce/vc-storefront/pull/487

PR #413, opened 2020-04-30
VP-2120: Fix incorrect totals calculation

ilyawzrd:
prevent sale price overwriting which leads to an incorrect totals calculation


yecli:
Released: https://github.com/VirtoCommerce/vc-storefront-core/releases/tag/v4.1.13

PR #415, opened 2020-05-08
VP-2194 Fix incorrect discount amount

AliveMen:
Get rid of double when work with money

### DIFF
--- a/VirtoCommerce.Storefront.Model/Common/Money/Money.cs
+++ b/VirtoCommerce.Storefront.Model/Common/Money/Money.cs
@@ -367,31 +367,6 @@ namespace VirtoCommerce.Storefront.Model.Common
             return result;
         }
 
-        /// <summary>
-        /// Evenly distributes the amount over n parts, resolving remainders that occur due to rounding 
-        /// errors, thereby garuanteeing the postcondition: result->sum(r|r.amount) = this.amount and
-        /// x elements in result are greater than at least one of the other elements, where x = amount mod n.
-        /// </summary>
-        /// <param name="n">Number of parts over which the amount is to be distibuted.</param>
-        /// <returns>Array with distributed Money amounts.</returns>
-        public Money[] Allocate(int n)
-        {
-            var cents = Math.Pow(10, DecimalDigits);
-            var lowResult = ((long)Math.Truncate((double)InternalAmount / n * cents)) / cents;
-            var highResult = lowResult + 1.0d / cents;
-            var remainder = (int)(((double)InternalAmount * cents) % n);
-
-            var results = new Money[n];
-
-            for (var i = 0; i < remainder; i++)
-                results[i] = new Money((decimal)highResult, Currency);
-
-            for (var i = remainder; i < n; i++)
-                results[i] = new Money((decimal)lowResult, Currency);
-
-            return results;
-        }
-
         #endregion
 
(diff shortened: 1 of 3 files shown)


VirtoCommerce.Storefront.Tests/Model/MoneyOperationTests.cs
```diff
@@ -37,5 +38,26 @@ public class MoneyOperationTests
             Assert.Equal(expectedFormattedAmountWithoutPoint, money.FormattedAmountWithoutPoint);
             Assert.Equal(expectedFormattedAmountWithoutPointAndCurrency, money.FormattedAmountWithoutPointAndCurrency);
         }
+
+        [Theory]
+        // Jira VP-2194 
+        [InlineData(4.6, 4.6, 1)]
```
yecli, line 44: I think here we should check specific array values.
4.6, 1 => [4.6]
10, 3 => [3.34, 3.33, 3.33]
All these values should be checked because otherwise it could be [0,0,10] - it is not what we want.

VirtoCommerce.Storefront.Tests/Model/MoneyOperationTests.cs
```diff
@@ -37,5 +38,26 @@ public class MoneyOperationTests
...
+        [InlineData(9.2, 9.2, 2)]
+        // Correct product allocation [3.34, 3.33, 3.33]
+        [InlineData(10, 10, 3)]
+        [InlineData(0.99, 0.99, 1)]
+        [InlineData(0.999, 0.999, 1)]
+        public void TestAllocate(decimal internalAmount, decimal expectedAmount, int quantity)
```
yecli, line 50: Need a meaningful name, like `Money_Allocate_DistributedCorrectly`

VirtoCommerce.Storefront.Tests/Model/MoneyOperationTests.cs
AliveMen: Changed in accordance with comments

VirtoCommerce.Storefront.Tests/Model/MoneyOperationTests.cs
AliveMen: changed

VirtoCommerce.Storefront.Model/Common/Money/Money.cs
```diff
@@ -369,25 +369,29 @@ public virtual string ToString(bool showDecimalDigits, bool showCurrencySymbol)
 
         /// <summary>
         /// Evenly distributes the amount over n parts, resolving remainders that occur due to rounding 
-        /// errors, thereby garuanteeing the postcondition: result->sum(r|r.amount) = this.amount and
+        /// errors, thereby garuanteeing the post condition: result->sum(r|r.amount) = this.amount and
         /// x elements in result are greater than at least one of the other elements, where x = amount mod n.
         /// </summary>
-        /// <param name="n">Number of parts over which the amount is to be distibuted.</param>
+        /// <param name="n">Number of parts over which the amount is to be distributed.</param>
```
yecli, line 375: As proposed by @tatarincev, need to remove `Allocate` method at all as redudant

AliveMen: done

PR #416, opened 2020-05-14
#VP-2081 add member id to identity result

kostyrin:


VirtoCommerce.Storefront.Model/Security/MemberIdIdentityResult.cs
```diff
@@ -0,0 +1,22 @@
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Identity;
+
+namespace VirtoCommerce.Storefront.Model.Security
+{
+    public class MemberIdIdentityResult : IdentityResult
```
tatarincev, line 7: I suggest defining a new type of `UserActionIdentityResult` instead of deriving from `IdentityResult` that will contain all properties from `IdentityResult` to preserving backward compatibility plus in addition with the new  `User` property.

t13ka:
When this pull request is going to be released?

PR #418, opened 2020-05-15
VP-2347: Implement IDictonary completely for MutablePagedList

trueboroda:
Implement Keys and Values properties of IDictionary


VirtoCommerce.Storefront.Model/Common/MutablePagedList.cs
```diff
@@ -238,14 +238,19 @@ public ICollection Keys
         {
             get
             {
-                throw new NotImplementedException();
+               TryReloadPagedData();
```
t13ka, line 241: The method 'TryReloadPagedData' tells us that there are only 2 ways of the results: true and false.
But we are using it like a void result. It's not the correct name of this method.

VirtoCommerce.Storefront.Model/Common/MutablePagedList.cs
```diff
@@ -238,14 +238,19 @@ public ICollection Keys
...
+               var dict = _pagedList.OfType<IAccessibleByIndexKey>().ToDictionary(x => x.IndexKey, x => x);
```
t13ka, line 242: .OfType can throw the null exception. So you should be careful with this operator in Generics.
For instance, the Breadcrumb is not implement the IAccessibleByIndexKey :)
I recommend to check the null reference and return something like this:
return Enumerable.Empty<string>().ToDictionary(x => x, x => x);

t13ka: the name 'dict' is not necessary should be short form in this context. Better name 'dictionary', it's more clear

t13ka: in the TryReloadPagedData method is located the hidden initialization of the _pagedList. It is wrong because you should to go and view this method to catch on the context

VirtoCommerce.Storefront/JsonConverters/BackwardCompatibilityJsonConverter.cs
```diff
@@ -46,7 +47,7 @@ public UserBackwardCompatibilityJsonConverter(JsonSerializerSettings jsonSetting
                 SerializationBinder = jsonSettings.SerializationBinder,
                 TypeNameAssemblyFormatHandling = jsonSettings.TypeNameAssemblyFormatHandling,
                 // exclude  BackCompatibilityJsonConverter from  Converters  to prevent infinite loops when serializing
-                Converters = null
+                Converters = jsonSettings.Converters.Except( jsonSettings.Converters.Where( x=> x.GetType() == typeof(UserBackwardCompatibilityJsonConverter)).ToList()).ToList()
```
tatarincev, line 50: redundant `Except(...ToList())`

VirtoCommerce.Storefront.Model/Common/MutablePagedList.cs
trueboroda: 'dict' is renamed to 'dictionary'.

PR #420, opened 2020-05-21
#VP-2335 FullName is required

kostyrin:


kostyrin:
@yecli actually there is fullname in db like required
image

VirtoCommerce.Storefront.Model/Security/UserRegistration.cs
```diff
@@ -10,12 +10,14 @@ public partial class UserRegistration
         public string PhotoUrl { get; set; }
 
         [FromForm(Name = "customer[first_name]")]
+        [Required]
```
tatarincev, line 13: I suggest marking these fields as required on the theme side in order to do not affect other themes where these fields are optional.

t13ka:
closed in favor of this https://github.com/VirtoCommerce/vc-theme-default/pull/150

PR #423, opened 2020-05-25
VP-2556: Implement an ability to manually set price, comment and dynamic properties of the cart line item

trueboroda:


VirtoCommerce.Storefront.Model/Cart/AddCartItem.cs
```diff
@@ -32,5 +33,24 @@ public string Id
         /// </summary>
         [JsonProperty("listName")]
         public string ListName { get; set; }
+
+        /// <summary>
+        /// Cart name
```
yecli, line 38: Need to fix comments

VirtoCommerce.Storefront.Model/Cart/Services/ICartBuilder.cs
```diff
@@ -50,7 +50,7 @@ public interface ICartBuilder
         /// <param name="product"></param>
         /// <param name="quantity"></param>
         /// <returns></returns>
-        Task<bool> AddItemAsync(Product product, int quantity);
+        Task<bool> AddItemAsync(Product product, int quantity, decimal? price = null, string comment = null);
```
tatarincev, line 53: need to replace a single parameter with `AddCartItem` type.

VirtoCommerce.Storefront.Model/Cart/Services/ICartBuilder.cs
```diff
@@ -165,5 +165,10 @@ public interface ICartBuilder
         Task ValidateAsync();
 
         Task SaveAsync();
+
+
+        Task ChangeItemPriceAsync(string lineItemId, decimal price);
```
tatarincev, line 170: Same as the commented above. Use `ChangeCartItemPrice` argument instead of multiple parameters.

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -578,17 +624,23 @@ protected virtual Task ChangeItemQuantityAsync(LineItem lineItem, int quantity)
...
-                    var salePrice = lineItem.Product.Price.GetTierPrice(quantity).Price;
-                    if (salePrice != 0)
-                    {
-                        lineItem.SalePrice = salePrice;
-                    }
-                    //List price should be always greater ot equals sale price because it may cause incorrect totals calculation
-                    if (lineItem.ListPrice < lineItem.SalePrice)
+                    var product = lineItem.Product;
+
+                    if (product.Price.ListPrice == lineItem.ListPrice)
```
tatarincev, line 632: This strange condition is required your clarification.

VirtoCommerce.Storefront.Model/Cart/AddCartItem.cs
```diff
@@ -32,5 +38,24 @@ public string Id
...
+        /// </summary>
+        [JsonProperty("comment")]
+        public string Comment { get; set; }
+
+        /// <summary>
+        /// Cart name
+        /// </summary>
+        [JsonProperty("dynamicProperties")]
+        public Dictionary<string, string> DynamicProperties { get; set; }
     }
```
trueboroda, line 60: There is no ability to turn off price validation.

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
trueboroda: This condition can be removed when 'custom price'  attribute will be added to the line item. this attribute means that price validation shouldn't need.

PR #424, opened 2020-05-27
VP-2407: Prevent organization registration with duplicated userName or email

ilyawzrd:
Right now an empty organization is being created despite errors.
We should check if there is a user with same UserName or Email before creating an organization.

### DIFF
--- a/VirtoCommerce.Storefront/Controllers/Api/ApiAccountController.cs
+++ b/VirtoCommerce.Storefront/Controllers/Api/ApiAccountController.cs
@@ -110,7 +110,12 @@ namespace VirtoCommerce.Storefront.Controllers.Api
 
             if (ModelState.IsValid)
             {
-                var user = await _userManager.FindByNameAsync(orgRegistration.UserName);
+                var user = await _userManager.FindByEmailAsync(orgRegistration.Email);
+                if (user != null)
+                {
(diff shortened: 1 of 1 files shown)


VirtoCommerce.Storefront/Controllers/Api/ApiAccountController.cs
```diff
@@ -128,10 +153,20 @@ public async Task<ActionResult<IdentityResult>> RegisterOrganization([FromBody]
                     await _signInManager.SignInAsync(user, isPersistent: true);
                     await _publisher.Publish(new UserLoginEvent(WorkContext, user));
                 }
+                else
```
tatarincev, line 156: redundant `else` block

VirtoCommerce.Storefront/Controllers/Api/ApiAccountController.cs
```diff
@@ -110,15 +110,40 @@ public async Task<ActionResult<IdentityResult>> RegisterOrganization([FromBody]
 
             if (ModelState.IsValid)
             {
+                var user = await _userManager.FindByEmailAsync(orgRegistration.Email);
```
tatarincev, line 113: Can you try to use `await userManager.UserValidators` to enforce more user validation rules?

PR #430, opened 2020-06-03
VP-2170: Fix contact & organization properties on registration

asvishnyakov:
* Add dynamic properties support to registration forms
* Fix missed object properties
* Allow use full user, contact, organization objects - for example, we want specify multiple phones for contact & organization


asvishnyakov:
@t13ka @tatarincev We use workaround for valueType (just hardcoded it in theme, it's possible because I use full Contact and Organization objects here), so no more blockers

asvishnyakov:
@tatarincev  It’s not only about dynamic properties. I added models of Contact & Organization because we wan to pass ANY existing property of Contact or Organization, for example phones or emails.

PR #440, opened 2020-06-10
Feature: add infrastructure for integration tests of cart api

YanBrovkin:
### Problem
During transition program logic with cart api interaction from Storefront to experience api module we have to control regression of functionality inside Storefront.

### Solution
It's necessary to create infrastructure for integration tests of cart api

### Proposed of changes
There are some infrastructure workaround & example test of reading empty cart


VirtoCommerce.Storefront.IntegrationTests/Cart/ReadingTests.cs
```diff
@@ -29,6 +71,29 @@ public void GetCartItemsCount_IfCartHasItems_ShouldReturnExactItemsCount()
         {
         }
 
+        private ShoppingCart GetNewCart()
```
Vectorfield4, line 74: Need to move this to "object builders" or register class with fixture in ctor of base class.

VirtoCommerce.Storefront.IntegrationTests/Infrastructure/MoneyJsonConverter.cs
```diff
@@ -0,0 +1,59 @@
...
+        public override bool CanRead => true;
+
+        public override Money ReadJson(JsonReader reader, Type objectType, Money existingValue, bool hasExistingValue, JsonSerializer serializer)
+        {
+            var obj = JObject.Load(reader);
+
+            var currencyCode = obj["currency"]["code"].Value<string>();
+            var amount = obj["amount"].Value<decimal>();
+
+            // TODO: here we should try to find incoming currency code inside available currencies,
```
Vectorfield4, line 21: What you think about removing this comments?

VirtoCommerce.Storefront.IntegrationTests/Cart/ReadingTests.cs
```diff
@@ -1,17 +1,59 @@
...
+                .Excluding(c => c.SelectedMemberPath == "Currency.CultureName")
+                .Excluding(c => c.SelectedMemberPath == "Currency.NumberFormat")
+                .Excluding(c => c.SelectedMemberPath == "Currency.EnglishName")
+                .Excluding(c => c.SelectedMemberInfo.Name == "CustomerId")
+                .Excluding(c => c.SelectedMemberInfo.Name == "ExtendedPriceTotal")
+                .Excluding(c => c.SelectedMemberInfo.Name == "ExtendedPriceTotalWithTax")
+                .ComparingByValue<ShoppingCart>()
+                .ComparingByMembers<User>()
+                .ComparingByMembers<Currency>()
+            );
```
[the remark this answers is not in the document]
YanBrovkin: removed

VirtoCommerce.Storefront.IntegrationTests/Cart/ReadingTests.cs
YanBrovkin: removed, unnecessary

VirtoCommerce.Storefront.IntegrationTests/Infrastructure/MoneyJsonConverter.cs
YanBrovkin: removed

PR #441, opened 2020-06-16
VP-3148: Fix issues after migration to Open API v3 and latest Swashbuckle

asvishnyakov:
### Problems
We have a few **problems** because of migration to Open API v3 and latest Swashbuckle:
1. POST methods body isn't named and has unknown position in generated code (backward incompatibility with Storefront v4 API clients)
2. POST methods body isn't named when scheme generated with Open API v2 specification (backward incompatibility with Storefront v4 API clients)
3. All reference and nullable types isn't marked as nullable in schema
4. POST methods body is required. ASP.NET Core will return 415 error code if you will send request with empty body to API with `[FromBody]` attribute
5. Some parameters can be nullable, but is required.
6. Storefront dosn't add Newtonsoft.Json support to Swashbuckle and instead used some custom filters

**Out of scope** of this PR:
2. Because of [bug](https://github.com/microsoft/OpenAPI.NET/issues/486) in OpenAPI.NET library which Swashbuckle use for API schema generation and conversion from v2 to v3. _The workaround_ for some tools is to use same vendor extensions (`x-` schema properties) as we use for v3 (see later). Extensibility of request body parameter in v2 also isn't supported currently, I made [PR](https://github.com/microsoft/OpenAPI.NET/pull/485) to OpenAPI.NET to support it.
5. I don't think it's something important just because it will make these parameters as union with `undefined` (`string | undefined`, for example). _It allow do more, than possible, not less._ So I created [VP-3175](https://virtocommerce.atlassian.net/browse/VP-3175) about that.

### Solution
1. Use vendor extensions to add naming and position to request body in schema:
1.1. `x-name` and `x-position` for `NSwag`
1.2 `x-codegen-request-body-name` for `Swagger Codegen`. Position isn't support and not neccessary here because Swagger Codegen always put body to the end of generated methods, even for Open API v2.
1.3. `x-ms-requestBody-name` and `x-ms-requestBody-index` for AutoRest.
2. Out of scope
3. Mark all reference and nullable types as nullable in schema
4. Mark all `requestBody` occurrences in schema as required
5. Out of scope
6. Properly configure Swashbuckle

### Proposed changes
1. I added `NameAndOrderRequestBody` parameter to `appSettings.json` in `Swagger:Schema` section. If it's enabled, `ParameterOrderFilter` and `NameAndOrderRequestBodyFilter` filers for Swashbuckle will add vendor extensions to support body naming and positionning.
https://github.com/VirtoCommerce/vc-storefront-core/pull/441/files#diff-3b8759d90ee0adf1adf31373f4c91f68R357-R358
https://github.com/VirtoCommerce/vc-storefront-core/pull/441/files#diff-94cddb6036b2fc6ff8d4aebebe23c31e
https://github.com/VirtoCommerce/vc-storefront-core/pull/441/files#diff-63affde52279701e934521b33cb972ef
2. The only improvement is `OpenApiSpecificationVersion` parameter in `Swagger:Schema` section of `appSettings.json` which allow you to select which schema version should be produced. Unfortunately, we can't add new API version with new Open API specification version support - all versions will be generated with the same Open API specification.
https://github.com/VirtoCommerce/vc-storefront-core/pull/441/files#diff-2f1bd9bddcbb37bce32c91dcbd59948e
https://github.com/VirtoCommerce/vc-storefront-core/pull/441/files#diff-bbef13df0bbb66cd265aff04d91e7476R9
https://github.com/VirtoCommerce/vc-storefront-core/pull/441/files#diff-3b8759d90ee0adf1adf31373f4c91f68R429-R434
3. Swashbuckle has a [bug](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1706) about that. Workaround is to enable `UseAllOfToExtendReferenceSchemas` schema option which will make desired behavior.
https://github.com/VirtoCommerce/vc-storefront-core/pull/441/files#diff-3b8759d90ee0adf1adf31373f4c91f68R367
4. `RequireRequestBodyFilter` added to make all `requestBody` occurrences in schema as required
https://github.com/VirtoCommerce/vc-storefront-core/pull/441/files#diff-3b8759d90ee0adf1adf31373f4c91f68R359
https://github.com/VirtoCommerce/vc-storefront-core/pull/441/files#diff-602d549f760206cd3d0f0882fcfe48ec
5. No changes
6. `AddSwaggerGenNewtonsoftSupport` method called during services configuration due to [Swashbuckle readme](https://github.com/domaindrivendev/Swashbuckle.AspNetCore#systemtextjson-stj-vs-newtonsoft). It also replace our `OptionalParametersFilter` and `NewtonsoftJsonIgnoreFilter` and produce more correct schema - for, example it exclude from schema proeprties marked with `[IgnoreDataMember]` like `Customer`, `Currency` or `Language` of `CartSearchCriteria`

### Additional context (optional)
1. NSwag has [bug](https://github.com/RicoSuter/NSwag/issues/2891), which prevent parameters ordering working right now, but I made [PR](https://github.com/RicoSuter/NSwag/pull/2905) to fix it.
2. NSwag uses NJsonScheme, which has a [bug](https://github.com/RicoSuter/NJsonSchema/issues/1197) about incorrect processing null values, but I made [PR](https://github.com/RicoSuter/NJsonSchema/pull/1210) to fix that
3. Swashbuckle use Microsoft OpenAPI.NET tool to generate API for v2, I created [bug](https://github.com/microsoft/OpenAPI.NET/issues/486) about support of this parameter and made [PR](https://github.com/microsoft/OpenAPI.NET/pull/485) to support extensions for body parameter - it's same workaround as for Open API v3.


lnetrebskii:
Review task https://virtocommerce.atlassian.net/browse/VP-3170 has been created

asvishnyakov:
I found issues with Newtonsoft.JSON configuration for Swashbuckle, so PR isn't complete, treat it as draft

Flanker72:
We don't need this anymore in newer vc-storefront releases.

PR #446, opened 2020-06-19
VP-3169 Add workflow to build docker image

mvktsk:
Added workflow which runs on push to meds/dev branch event and send an event to
VirtoCommerce/vc-meds-docker for rebuild meds-storefront docker image

### DIFF
--- a/.github/workflows/meds.yml
+++ b/.github/workflows/meds.yml
@@ -1,0 +1,26 @@
+name: MEDs docker
+
+# Controls when the action will run. Triggers the workflow on release 
+# event
+
+on:
+  push:
+    branches: [meds\dev]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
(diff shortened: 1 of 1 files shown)


.github/workflows/meds.yml
```diff
@@ -0,0 +1,26 @@
...
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+
+  webhook:
+    name: Send event
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set platform repository mame
```
lnetrebskii, line 18: You can simplify the environment variable declaration like here https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#about-environment-variables

mvktsk: Fixed

PR #456, opened 2020-07-02
#VP-3470 add integration tests for AccountController

kostyrin:


VirtoCommerce.Storefront.IntegrationTests/Tests/ApiAccountControllerTests.cs
```diff
@@ -0,0 +1,205 @@
...
+            var content = new StringContent(
+                JsonConvert.SerializeObject(organization),
+                Encoding.UTF8,
+                "application/json");
+
+            //Act
+            await _client.PutAsync(TestEnvironment.OrganizationEndpoint, content);
+
+            //Assert
+            _client.Logout();
```
eMazeika, line 99: can check the name was changed?
Also, should test address changing

kostyrin: it's temporally skipped. there is a problem with permissions. need to resolve it

PR #471, opened 2020-07-14
VP-3405: Add support of global settings

asvishnyakov:
All settings outside presets are treated as global and loaded into every preset.

VP-3402 also should be fixed by this PR

### DIFF
--- a/VirtoCommerce.LiquidThemeEngine/SettingsManager.cs
+++ b/VirtoCommerce.LiquidThemeEngine/SettingsManager.cs
@@ -70,20 +70,35 @@ namespace VirtoCommerce.LiquidThemeEngine
             }
 
             var currentPresetJsonToken = json.GetValue("current");
-            if (currentPresetJsonToken is JValue currentPresetJsonValue)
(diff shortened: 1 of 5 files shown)


VirtoCommerce.LiquidThemeEngine/SettingsManager.cs
```diff
@@ -70,20 +70,35 @@ public static Settings ReadSettings(JObject json)
...
             {
-                result.CurrentPreset.Json = currentPresetJsonToken as JObject;
+                var globalSettings = json.Properties().Where(x => x.Name != "presets" && x.Name != "current");
+                foreach (var globalSetting in globalSettings)
+                {
+                    foreach (var preset in result.Presets)
+                    {
+                        preset.Json.Add(globalSetting.Name, globalSetting.Value);
+                    }
+                }
```
t13ka, line 101: It would be nice to see the new unit tests concerned this logic

t13ka:
to be continued..

Flanker72:
We don't need this anymore in newer vc-storefront releases.

PR #479, opened 2020-07-24
VME-812

blazemachine:
### Problem
We need to create possibility to send requests to Azure Logic Apps. Also not recommended keep URLs to them apps in theme.

### Solution
To send request to middleware with protection and to hide URLs of Azure Logic Apps.

### Proposed of changes
Was created controller with injection of services that contains logic of extract information (URL) about Azure Logic Apps from configuration. Controller is using anti forgery token for protection.


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-3793 has been created

basilkot:
wrong target repository

PR #484, opened 2020-07-28
VP-3582: Add outlines property to product for promotion evaluation

ilyawzrd:
### Problem
Marketing promotions don’t work as expected, because the wrong one outline for product is chosen.

### Solution
In DynamicExpressionsModule we have support for multiple outlines, so just need to pass string that contains multiple outlines with separator 

### Proposed of changes
Add new Product propetry - Outlines, that contains all outlines of product with separator and then pass it to ProductPromoEntry

### DIFF
--- a/VirtoCommerce.Storefront.Model/Catalog/Product.cs
+++ b/VirtoCommerce.Storefront.Model/Catalog/Product.cs
@@ -72,6 +72,10 @@ namespace VirtoCommerce.Storefront.Model.Catalog
         /// All parent categories ids concatenated with "/". E.g. (1/21/344)
         /// </summary>
         public string Outline { get; set; }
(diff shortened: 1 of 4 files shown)


VirtoCommerce.Storefront.Model/Catalog/Product.cs
```diff
@@ -72,6 +72,10 @@ public Product(Currency currency, Language language)
         /// All parent categories ids concatenated with "/". E.g. (1/21/344)
         /// </summary>
         public string Outline { get; set; }
+        ///// <summary>
+        /// All outlines of product
```
yecli, line 76: Need to be:
All relative product outlines for the given catalog

PR #491, opened 2020-08-03
Fix ci workflow sonar error

mvktsk:


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-3953 has been created

.github/workflows/main.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       run: vc-build Compile
 
     - name: Unit Tests
-      run: vc-build Test -TestsFilter "Category=Unit|Category=CI" -skip Restore+Compile
+      run: vc-build Test -TestsFilter "Category=Unit|Category=CI" -skip
```
yecli, line 57: Do we need "-skip" param name if no values presented for it?

mvktsk: Yes, without it, all targets on which Tests depend will run

PR #500, opened 2020-08-07
VP-3777: Hide non-priced products from store

akak1977:


Woland2k:
Non priced items should still appear, we have a checkbox saying do not show in storefront which is used to hide items. The non priced products shouldn't be available to purchase however.

akak1977:
Well, take a look please, yet another time.
Non-priced items appearing as unavailable.

VirtoCommerce.Storefront/Domain/Pricing/PricingService.cs
```diff
@@ -114,6 +110,7 @@ protected virtual void ApplyProductPrices(IEnumerable<Product> products, IList<p
                 var productPrices = prices.Where(x => x.ProductId == product.Id)
                                           .Select(x => x.ToProductPrice(workContext.AllCurrencies, workContext.CurrentLanguage));
                 product.ApplyPrices(productPrices, workContext.CurrentCurrency, workContext.AllCurrencies);
+                product.IsBuyable &= productPrices.Any();
```
tatarincev, line 113: you mustn't set `IsBuyable`  directly add such logic to the `ProductIsBuyableSpecification` instead see how this done https://github.com/VirtoCommerce/vc-module-experience-api/blob/dev/src/XPurchase/VirtoCommerce.XPurchase/Specifications/ProductIsBuyableSpecification.cs#L12

akak1977: It would be good, but there is no way to know prices is empty after ApplyPrices call. After call all the Prices will filled up with default instances (not null Price and non-empty Prices collection).

PR #503, opened 2020-08-10
VP-1888: Fix bug: empty inventory causes ability to buy unlimited items

akak1977:


Woland2k:
TrackInventory is just for tracking and not enforcing inventory. So should be able to oder any number of items just making sure inventory is tracked.

PR #506, opened 2020-08-11
VP-4070: Update Autorest clients

eMazeika:
VP-4070: https://virtocommerce.atlassian.net/browse/VP-4070


akak1977:
This does not need

PR #508, opened 2020-08-12
VP-3968: Fix error when user add more products than contains in stock

AliveMen:
### Problem
error when user add more products than contains in stock

### Solution
Extend error model received from x-api module
Fill lineItems.validationErrors list
Minor fixes


VirtoCommerce.Storefront.Model/Cart/ValidationErrors/ValidationError.cs
```diff
@@ -4,11 +4,9 @@ namespace VirtoCommerce.Storefront.Model.Cart.ValidationErrors
 {
     public abstract class ValidationError : ValueObject
     {
-        protected ValidationError()
```
yecli, line 7: Why this constructor was removed?

PR #510, opened 2020-08-12
VP-3971: Fix associations for product page

Vectorfield4:
https://virtocommerce.atlassian.net/browse/VP-3971

### Proposed of changes
Describe the technical details of realization provided by you.

- [ ] Check all the changes in github PR - files count (non of them are redundant, have meaningful changes, all are added), if target branch is correct
- [ ] Check methods and variable namings - it should be self descriptive, no typos
- [ ] Check you did not introduce breaking changes in API and public models/services.
- [ ] Respect extensibility - https://community.virtocommerce.com/t/extensibility-basics-the-domain-model-and-persistence-layer-extension/141
- [ ] Follow [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and [SOLID](https://en.wikipedia.org/wiki/SOLID) principles
- [ ] For unit tests - follow Microsoft best practices: https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices
- [ ] Consolidate solution dependencies in case you are using newer version, update VC module dependencies in module.manifest. Do not upgrade 3rd party packages that are shipped with the platform with the version newer than in the platform.
- [ ] Check code style conventions - https://github.com/VirtoCommerce/styleguide/blob/master/csharp.md
- [ ] Check PR have a concise and descriptive title, follow [git commit message rules](https://github.com/VirtoCommerce/styleguide/blob/master/gitcommits.md)


VirtoCommerce.Storefront/Domain/Catalog/ExperienceCatalogService.cs
```diff
@@ -119,44 +122,51 @@ public virtual async Task<CatalogSearchResult> SearchProductsAsync(ProductSearch
...
-                {
-                    association.Product = allAssociations.FirstOrDefault(x => x.Id == association.ProductId);
+                return;
+            }
 
-                    if (association.Product != null)
-                    {
-                        EstablishLazyDependenciesForProducts(new[] { association.Product });
-                    }
+            var allAssociations = await GetProductsAsync(productAssociations.Select(x => x.Product.Id).ToArray());
```
yecli, line 159: Why `x.Product.Id` is used instead of `x.ProductId`? Is `ProductId` empty?

VirtoCommerce.Storefront.Model/Catalog/Services/ICatalogSearchService.cs
```diff
@@ -21,5 +22,7 @@ public interface ICatalogService
         CatalogSearchResult SearchProducts(ProductSearchCriteria criteria);
 
         IPagedList<Category> SearchCategories(CategorySearchCriteria criteria);
+
+        Task LoadAssociations(ProductAssociation[] productAssociations);
```
yecli, line 26: Why do we need this very specific method in the interface?
I think we need to load associations based on responseGroup in both GetProductsAsync and SearchProductsAsync

VirtoCommerce.Storefront/Domain/Catalog/ExperienceCatalogService.cs
Vectorfield4: It contains the wrong id for associations. ProductId is the main id of the product, so we cant get associations with this id.

VirtoCommerce.Storefront.Model/Catalog/Services/ICatalogSearchService.cs
Vectorfield4: We don't use response groups when we working with expApi and current realization load only association ids in product object to lead "lazy loading" pattern, so we need an additional method to call it outside and inside service.

PR #528, opened 2020-08-24
VP-1575 Fix 400 error when reset password

tatarincev:
Manually set User in the HttpContext to avoid issues like the Antiforgery  cookies are  generated for an anonymous user within this request despite that fact what  user has signed in already


eMazeika (approved):

eMazeika:
@tatarincev , also check my changes

akak1977 (approved):

PR #532, opened 2020-08-28
VP-2076: Preview of theme always open active theme

pushnitsa:
### Problem
There is no way to preview any other theme besides active.

### Solution
Theme should be able to preview using query string **?previewtheme=**

### Proposed of changes
Add query string interceptor and replace theme name by passed from **previewtheme** parameter.

### Additional context (optional)
theme_preview

### DIFF
--- a/VirtoCommerce.Storefront/Domain/Stores/StoreWorkContextBuilderExtensions.cs
+++ b/VirtoCommerce.Storefront/Domain/Stores/StoreWorkContextBuilderExtensions.cs
@@ -1,9 +1,9 @@
-using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


AliveMen (approved):

VirtoCommerce.Storefront/Domain/Stores/StoreWorkContextBuilderExtensions.cs
```diff
@@ -22,6 +22,9 @@ public static Task WithStoresAsync(this IWorkContextBuilder builder, IEnumerable
 
             builder.WorkContext.AllStores = stores.ToArray();
             var currentStore = builder.HttpContext.GetCurrentStore(stores, defaultStoreId);
+
+            currentStore.ReplaceThemeForPreviewIfRequired(builder.HttpContext);
```
tatarincev, line 26: I would like to simplify implementation and replace   all your code with this single line
`currentStore.ThemeName  = builder.WorkContext.QueryString.Get("previewtheme") ?? currentStore.ThemeName;`

AliveMen (approved):

tatarincev (approved):

PR #536, opened 2020-09-01
VP-3949: Return payment method name instead of typeName

ilyawzrd:
Fix for VP-3949: We need to show payment method name instead of typeName
normalPaymentName


OlegoO (approved):

VirtoCommerce.Storefront/Domain/Stores/StoreService.cs
```diff
@@ -80,7 +80,7 @@ protected virtual async Task<Store> ConvertStoreAndLoadDependeciesAsync(StoreApi
                     .Select(pm =>
                     {
                         var paymentMethod = pm.ToStorePaymentMethod(currency);
-                        paymentMethod.Name = pm.TypeName;
+                        paymentMethod.Name = pm.Name;
```
tatarincev, line 83: This is breaking changes for other themes,  I suggest adding a new field `DisplayName`  and left the old `Name` as is.

PR #538, opened 2020-09-02
VP-960: Fix bug: unable to delete product from the list after adding all products to the cart

akak1977:


AliveMen (dismissed):

akak1977:
It seems we should do the same for shipments and payments (Clone and id=null)

VirtoCommerce.Storefront/Domain/Cart/CartBuilder.cs
```diff
@@ -332,17 +332,11 @@ public virtual async Task MergeWithCartAsync(ShoppingCart cart)
...
-            //To prevent insertions same Ids for target cart
-            //exclude user because it might be the current one
-            var entities = cart.GetFlatObjectsListWithInterface<IEntity>();
-            foreach (var entity in entities.Where(x => !(x is User)).ToList())
-            {
-                entity.Id = null;
-            }
 
-            foreach (var lineItem in cart.Items)
+
```
tatarincev, line 336: You need to reset ids for the cloned original cart instance.  Just add `cart.Clone()` before ids resetting.

akak1977: OK

PR #551, opened 2020-09-23
Use aggregation field instead of label

sjroesink:
Use aggregation field instead of label when building term url

### Problem
When building the term url, the label of an aggregation is used instead of the name.


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-5091 has been created

tatarincev (dismissed):

PR #559, opened 2020-11-19
VP-4321: Add additional headers to every AuthHandler

akak1977:


VirtoCommerce.Storefront/Infrastructure/Autorest/BaseAuthHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+        private readonly IWorkContextAccessor _workContextAccessor;
+        private readonly IHttpContextAccessor _httpContextAccessor;
+
+        public BaseAuthHandler(IWorkContextAccessor workContextAccessor, IHttpContextAccessor httpContextAccessor)
+        {
+            _workContextAccessor = workContextAccessor;
+            _httpContextAccessor = httpContextAccessor;
+        }
+
+        protected void AddCurrentUser(HttpRequestMessage request)
```
yecli, line 18: Maybe make this class abstract, add `AddPlaformAuth` abstract method that should be implemented in the descendants, override `SendAsync` here with these lines
```
            AddCurrentUser(request);
            AddUserIp(request);

            AddPlaformAuth(request);
```
That there will be no code duplication and the addition of current user could not be forgotten in the descendants

tatarincev (approved):

yecli (approved):

PR #562, opened 2021-01-18
ValueObjects with different values should not be equal and should have different cache keys

artem-dudarev:
This PR fixes the following issues that lead to incorrect equality components and duplicate cache keys:
1. `null` and `"null"` values are considered equal.
2. Empty collections are just ignored.

### DIFF
--- a/VirtoCommerce.Storefront.Model/Common/ValueObject.cs
+++ b/VirtoCommerce.Storefront.Model/Common/ValueObject.cs
@@ -4,7 +4,6 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Web;
 
 namespace VirtoCommerce.Storefront.Model.Common
 {
@@ -49,7 +48,11 @@ namespace VirtoCommerce.Storefront.Model.Common
 
         public virtual string GetCacheKey()
         {
-            return string.Join("|", GetEqualityComponents().Select(x => x ?? "null").Select(x => x is ICacheKey cacheKey ? cacheKey.GetCacheKey() : x.ToString()));
+            var keyValues = GetEqualityComponents()
+                .Select(x => x is string ? $"'{x}'" : x)
+                .Select(x => x is ICacheKey cacheKey ? cacheKey.GetCacheKey() : x?.ToString());
+
+            return string.Join("|", keyValues);
         }
 
         protected virtual IEnumerable<object> GetEqualityComponents()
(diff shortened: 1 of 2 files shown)


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-6667 has been created

VirtoCommerce.Storefront.Model/Common/ValueObject.cs
```diff
@@ -59,17 +62,19 @@ protected virtual IEnumerable<object> GetEqualityComponents()
                 var value = property.GetValue(this);
                 if (value == null)
                 {
-                    yield return "null";
+                    yield return null;
```
tatarincev, line 65: it was important to return "null" as a string if the value not set to preserve properties order when a cache key constructing. The `string.Join`  will skip null values and the resulting cache key may contain a various number of key parts.

artem-dudarev: `null` and `"null"` are different values. When you replace one with another you get the same equality components for different values and you get duplicate cache keys. This PR fixes exactly this bug.

artem-dudarev: As for _preserving  properties order_ - this order is broken when you just skip empty collections, and two objects
`{ List1 = { "A" }, List2 = { } }`
`{ List1 = { }, List2 = { "A" } }`
become equal and produce the same cache key. This bug is also fixed in this PR.

artem-dudarev: All these cases are covered by tests.

artem-dudarev: BTW, `string.Join()` doesn't skip `null` values. `string.Join("|", null, null, null)` returns `||`.

tatarincev (approved):

VirtoCommerce.Storefront.Model/Common/ValueObject.cs
tatarincev: Could you do the same PR for the platform?
https://github.com/VirtoCommerce/vc-platform/blob/master/src/VirtoCommerce.Platform.Core/Common/ValueObject.cs

artem-dudarev: Sure

tatarincev (approved):

OlegoO (approved):

PR #570, opened 2021-03-01
VP-6887: Set verified to false when changed email address in the existent account

AliveMen:
Set emailConfirmed to false if user change own email address


ksavosteev (approved):

Vectorfield4 (approved):

eMazeika:
@Vectorfield4 this change was redundant, as the identical checks are also performed in Platform: https://github.com/VirtoCommerce/vc-platform/blob/dev/src/VirtoCommerce.Platform.Web/Controllers/Api/SecurityController.cs#L546
I'd suggest to revert.

Vectorfield4:
@eMazeika thanks for suggesting. I think you are right, but reverting "changes" may bring some risks.

PR #577, opened 2021-07-22
PT-1920 added check for graphql auth error

ksavosteev:
Example on how to check graphql auth errors

### DIFF
--- a/VirtoCommerce.Storefront/Infrastructure/GraphQLExecutionResult.cs
+++ b/VirtoCommerce.Storefront/Infrastructure/GraphQLExecutionResult.cs
@@ -1,0 +1,17 @@
+namespace VirtoCommerce.Storefront.Infrastructure
+{
+    public class GraphQLExecutionResult
+    {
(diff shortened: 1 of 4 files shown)


VirtoCommerce.Storefront/Infrastructure/GraphQLExecutionResult.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    public class GraphQLError
+    {
+        public GraphQLExtensions Extensions { get; set; }
+    }
+
+    public class GraphQLExtensions
+    {
+        public string Code { get; set; }
+    }
```
Vectorfield4, line 16:
```suggestion
    public class GraphQLExecutionResult
    {
        public GraphQLError[] Errors { get; set; }
        
        public class GraphQLError
        {
            public GraphQLExtensions Extensions { get; set; }
            
            public class GraphQLExtensions
            {
                public string Code { get; set; }
            }
        }
    }
```

PR #579, opened 2021-10-08
PT-4940 Custom notifications and file uploading through endpoint in Storefront

alexeyshibanov:
Added API endpoint in storefront to recieve multipart form data, send files (if exists) to platform blob storage and send notification with links to files


Flanker72:
We are not planning to include such functionality in the default vc-storefront codebase. You may implement this scenario in the particular project if needed.

PR #595, opened 2022-02-10
VME-1331: Fix metadata reading for content loader. Update for empty json arrays.

blazemachine:
https://virtocommerce.atlassian.net/browse/VME-1331


Flanker72:
We don't need this anymore in newer vc-storefront releases. Such logic may corrupt the existing projects. Also you haven't provided any accessable description about what's this. I see, some files are related directly to virtocommerce.com website - it's definitely not what we want in the main codebase.

PR #601, opened 2022-04-27
We should check and get the defaultStoreId first

doanson44:
In case we config default store id in setting file, we should check the condition and get store by defaultStoreId before fallback to get default store.


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-8019 has been created

PR #642, opened 2023-04-12
Revert "feat: add health check to connection to platform"

basilkot:
Reverts VirtoCommerce/vc-storefront#641


basilkot:
reverted by mistake

PR #665, opened 2024-01-16
Switched to AsyncKeyedLock

MarkCiliaVincenti:
Fixes an issue in MemoryCacheExtensions where GetOrCreateExclusiveAsync and GetOrCreateExclusive were not locking on the same lock and could run concurrently. Furthermore, two concurrent dictionaries were being filled with lock objects that were not being cleared from memory.


OlegoO:
Thank you for reaching out to us with this contribution. Virto Commerce Storefront is going to be deprecated and archived in the next few months.  At this moment, we didn't approve any new contributions.

